PR #34, opened 2018-09-04
Fixing styles imports and button component when running 'npm run dev'

rssilva:
I'm seeing these problems when running `npm run dev`

screen shot 2018-09-04 at 4 53 33 pm

screen shot 2018-09-04 at 4 54 17 pm


tassoevan:
@ggazzo The first bug is an annoying issue from https://github.com/jenseng/react-i18nliner/issues/34 and it doesn't seem to be solved soon.

PR #88, opened 2019-01-10
Agent's Custom Field added in storybook

### LINKED ISSUE #72: Support Agent's custom fields
In our current version of Livechat, we display data from the agent's custom fields, such as 'Phone Number', etc.
This data is displayed in the contents of the header, as we can see below:

screen shot 2018-11-29 at 17 13 41
 
We need to support it in our new Livechat client.

- [ ] apply on storybook

madhavrathi:
Fixes #72 
Screenshot:
screenshot 2019-01-11 at 12 43 34 am
@renatobecker Please Review.
Thanks


renatobecker:
@madhavrathi, I've reviewed your PR.
I noticed that following the sketch, when the `custom` component is displayed, the Headers' height needs to be `height: 77px;` instead of the default height: `height: 56px;`.

But let's wait the review from @tassoevan, he's our front-end expert guy!

Thanks in advance.

src/components/Screen/index.js
```diff
@@ -71,6 +71,11 @@ export class Screen extends Component {
 							<span>{agent.email}</span>
 						</Header.SubTitle>
 					)}
+					{agent.phone && (
```
renatobecker: @madhavrathi, you need to check if the agent property is assigned before reading the `agent.phone`:

`agent && agent.phone`.

tassoevan:
@renatobecker *70px
Also, the Avatar component must be adjusted for this size, maybe adapting its size based on parent's height. I think the current sizes must be updated -- we have documented only 46x46 and 36x36, but @brunosquadros didn't design the typing indicator, which uses 32x32.

PR #105, opened 2019-01-16
Add new alert component when chat transfer fails

renatobecker:
New alert added when the transfer chat went wrong.


src/routes/SwitchDepartment/container.js
```diff
@@ -31,24 +31,20 @@ export class SwitchDepartmentContainer extends Component {
 			const result = await SDK.transferChat({ rid, department });
 			const { success } = result;
 			if (!success) {
-				return ModalManager.alert({
-					text: I18n.t('No available agents to transfer'),
-				});
+				await dispatch({ alerts: insert(alerts, { id: createToken(), children: I18n.t('No available agents to transfer'), warning: true, timeout: 5000 }) });
```
ggazzo: why not throw an error  and remove this line? I mean, I'm not sure if we should continue this logic if we got a success == false

renatobecker: @ggazzo, can you please review it again?

PR #120, opened 2019-01-22
Show composer actions only if it has text

### LINKED ISSUE #96: toggle 'icon send' and 'add file' if message compose has text or not

diegolmello:
Closes #96 

jan-22-2019 15-40-35

### DIFF
--- a/src/components/Composer/index.js
+++ b/src/components/Composer/index.js
@@ -21,13 +21,41 @@ export const Action = ({ children, onClick, ...args }) => (
 	>{children}</button>
 );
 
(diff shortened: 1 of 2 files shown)


diegolmello:
@renatobecker @ggazzo @tassoevan Can you guys help reviewing this one?
It seems easy, but it has a lot of workarounds to pass through.
We can't rerender the composer every time, because it would lose track of cursor/caret.

Source:
https://stackoverflow.com/questions/22677931/react-js-onchange-event-for-contenteditable/27255103#27255103

PR #124, opened 2019-01-23
Set standard scoped package name

tassoevan:


package.json
```diff
@@ -1,5 +1,5 @@
 {
-	"name": "rocket.chat.livechat",
+	"name": "@rocket.chat/livechat",
```
renatobecker: It would be better naming this package as `enterprise` because we have plans to have a community version in a near feature.
`@rocket.chat/livechat-enterprise` or `@rocket.chat/livechat.enterprise`... something like this.

PR #135, opened 2019-01-24
Get server URL from __meteor_runtime_config__ too

tassoevan:


src/api.js
```diff
@@ -1,9 +1,10 @@
 import queryString from 'query-string';
 import LivechatClient from '@rocket.chat/sdk/lib/clients/Livechat';
 
+const host = (
```
renatobecker: What about injecting a global named as `ROOT_URL` instead of passing `__meteor_runtime_config__`?
It seems like we have a `meteor` dependency, don't you think?

ggazzo: I really like the idea to remove any mention to `meteor` in this layer of the project...

PR #148, opened 2019-02-01
[NEW] Support pt-BR language

renatobecker:


tassoevan:
@renatobecker I just some missing strings. Could you check them?

renatobecker:
Hi @tassoevan, everything is looking good, but we're still facing an issue related to the language name that is being served by the `config` endpoint, I mean, the server is sending `pt-br` string but the client is expecting `pt-BR` string. Can we fix it before approving this PR?

Thanks.

PR #157, opened 2019-02-12
Fix #91 Better tooltip

### LINKED ISSUE #91: Better Tooltip
- [ ] option no animate 
- [ ] improve the calc to define the arrow/container position if the tooltip pass the page boundaries 

image

jervis446:
This PR resolves #91 better tooltip.

`Present behavior`
1)
screenshot from 2019-02-12 13-14-19

2)
screenshot from 2019-02-12 13-16-04 1

I've checked the Z-index value Manually and it worked fine. I have doubt about the expected behaviour tooltip position of arrow.

is it this one.
screenshot from 2019-02-12 13-16-04 1

or this one.
51048465-23c53700-15b2-11e9-807b-fbc743770925


jervis446:
@ggazzo PTAL

ggazzo:
@jervis446 sorry , but I couldnt realise the difference between old and new code.

about the arrow positions there are two situations:

first: there are no more space left, so the tooltip needs find another place to open, the arrow should keep where the anchor/trigger is.
second: there are space, so the arrow and the tooltip get centered by the anchor/trigger

jervis446:
@ggazzo Sorry for not reply, I was busy due to mid-sem exams... I'll continue work from tomorrow :)

jervis446:
> @jervis446 sorry , but I couldnt realise the difference between old and new code.
> 
> about the arrow positions there are two situations:
> 
> first: there are no more space left, so the tooltip needs find another place to open, the arrow should keep where the anchor/trigger is.
> second: there are space, so the arrow and the tooltip get centered by the anchor/trigger

So means I have to make changes accordingly? but how do I test the second condition on my server because in my server there is always no space available.

jervis446:
Hi, @ggazzo 

image

this is good? , when there is space is available.
And one more thing for testing, I have to use storybook?

jervis446:
Hi @ggazzo,Now please PTAL

jervis446:
I'll resolve branch conflicts tomorrow, but you can review it by download this branch in your server :)

package.json
```diff
@@ -69,6 +69,7 @@
 		"preact-i18nline": "^2.0.0",
 		"preact-router": "^2.5.7",
 		"query-string": "6.2.0",
+		"react-responsive": "^6.1.1",
```
tassoevan: We're not using React in this project. `preact-compat` is applied only for compatibility with Storybook.

jervis446: Okay, I'll change some alternative :)

jervis446: hey @tassoevan can I use preact-fluid ?
it also has grid system.

jervis446: let me implement and test it.

tassoevan: I don't recommend using preact-fluid because we're doing here from scratch exactly what preact-fluid is: a UI toolkit. Also, I don't think the tooltips on preact-fluid calculate boundaries to reposition themselves, one of requirements of #157.

jervis446: Hey @tassoevan 
Sorry for the late reply I was busy with my Midsem exams,
Ya, actually yesterday I was figuring out the way to do media queries through preact and react but not able to find the way from scratch. can I use CSS plugin for that or please suggest the way :)

PR #163, opened 2019-02-22
[FIX] Get next agent

renatobecker:
The department field stored over the iframe API was not being passed over the `Livechat.nextAgent` Method.


src/lib/hooks.js
```diff
@@ -72,8 +72,8 @@ const api = {
 	},
 
 	clearDepartment() {
-		const { iframe } = store.state;
-		store.setState({ iframe: { ...iframe, department: null } });
+		const { iframe, iframe: { guest } } = store.state;
```
tassoevan:
```suggestion
		const { iframe: { guest, ...iframe } } = store.state;
```

PR #169, opened 2019-02-28
[IMPROVE] Dismiss Popover with escape key

### LINKED ISSUE #143: Dismiss modal/popover on 'esc'

Aditya-Kolla:
This is a WIP PR. 
I added a `esc` key listener to the popover component which will track the keydown. 
As of now, it works fine for the 'options' popover.

fixes #143


Aditya-Kolla:
@renatobecker As of now, the esc key will work for popovers. 
Could you help me figure out the same for Modals?
The onDismiss() method needs a target id if I'm right?
How should I approach this

ggazzo:
@tassoevan

PR #171, opened 2019-02-28
[FIX] Agent phone number

renatobecker:
Some times the agent's phone number is defined on `agent.phone` array and this case wasn't been expected on the widget.


renatobecker:
Yeah, I agree.
I'll improve the agent's data structure, we have the same case with the e-mail field, so I'll take a look at it later.

PR #177, opened 2019-03-07
Fix setDepartment method

### LINKED ISSUE #175: `setDepartment` method is not supporting department name as parameter
The `setDepartment` method is not working when the `department name` is passed by parameter.

renatobecker:
Closes #175 

The `setDepartment` method wasn't supporting to pass the name of the department as a parameter.


src/lib/hooks.js
```diff
@@ -59,21 +59,17 @@ const api = {
...
+		const dept = departments.find((dep) => dep._id === value || dep.name === value);
+		const department = (dept && dept._id) || '';
 
 		updateIframeGuestData({ department });
 	},
 
 	clearDepartment() {
-		const { iframe: { guest, ...iframe } } = store.state;
-		store.setState({ iframe: { ...iframe, guest: { ...guest, department: null } } });
+		updateIframeGuestData({ department: '' });
```
tassoevan: Why an empty string instead of `null`?

renatobecker: Yeah, I know about that.. It's another case related to the SDK methods, We'll need to work on the SDK and then we'll be able to pass `null` instead of `string`.

PR #203, opened 2019-03-19
Modal background color

### LINKED ISSUE #192: make the fade of the background dark, as for the popups
The current white fade makes the model blend with the rest on the top

image

renatobecker:
Closes #192.

Changed the `overlay background color` of the Modal component, applying the same color used in the popup.

Screen Shot 2019-03-19 at 09 46 46

### DIFF
--- a/src/components/Menu/styles.scss
+++ b/src/components/Menu/styles.scss
@@ -128,5 +128,5 @@
 }
 
 .popover-menu__overlay {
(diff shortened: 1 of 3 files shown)


renatobecker:
> https://github.com/RocketChat/Rocket.Chat.Livechat/blob/dev/src/components/Menu/styles.scss#L131
> 
> I think we should use a variable to keep all overlays same

@ggazzo, done.
Thanks.

PR #210, opened 2019-03-29
Alert improvements

### LINKED ISSUE #193: Allow users to send messages while they are in the queue
It is very annoying not being able to send a message while waiting to be connected.

image

renatobecker:
Closes #193.

This PR removes the `connecting agent` message from `Composer` component.
Now, it is displayed as a regular alert as requested in the issue mentioned above.

Screen Shot 2019-03-29 at 14 52 02

Other improvements are related to connecting/disconnecting stuff. When the client is disconnected, an alert will be displayed on the top of the widget, like shown below:

Screen Shot 2019-04-03 at 11 13 05

If the widget gets connected, an alert will be displayed too, as well as the widget will subscribe the room and load the messages again.


src/components/Alert/index.js
```diff
@@ -7,6 +7,8 @@ import styles from './styles';
 export class Alert extends Component {
 	static defaultProps = {
 		timeout: 3000,
+		autoClose: true,
```
tassoevan: Prefer `timeout={0}` over introducing an falsifiable `autoClose` property.

renatobecker: Fixed.

PR #214, opened 2019-04-16
[NEW] Add setLanguage method in livechat api methods

knrt10:
TODO
- [x] Store language in `iframe`
- [x] Create a new file for language handling and handle all the task there

### DIFF
--- a/src/components/App/index.js
+++ b/src/components/App/index.js
@@ -5,6 +5,7 @@ import { Livechat } from '../../api';
 import history from '../../history';
 import { loadConfig, clearConnectionAlerts } from '../../lib/main';
 import CustomFields from '../../lib/customFields';
+import Language from '../../lib/language';
 import Triggers from '../../lib/triggers';
 import Hooks from '../../lib/hooks';
 import { parentCall } from '../../lib/parentCall';
@@ -146,6 +147,7 @@ export class App extends Component {
 		await Livechat.connect();
 		this.handleTriggers();
 		CustomFields.init();
+		Language.init();
 		Hooks.init();
 		userPresence.init();
 
@@ -166,25 +168,12 @@ export class App extends Component {
 			return language;
 		};
 
-		const browserLanguage = () => (navigator.userLanguage || navigator.language);
-
-		const normalizeLanguageString = (languageString) => {
-			let [languageCode, countryCode] = languageString.split ? languageString.split(/[-_]/) : [];
-			if (!languageCode || languageCode.length !== 2) {
-				return 'en';
-			}
-			languageCode = languageCode.toLowerCase();
-
-			if (!countryCode || countryCode.length !== 2) {
-				countryCode = null;
-			} else {
-				countryCode = countryCode.toUpperCase();
-			}
-
-			return countryCode ? `${ languageCode }_${ countryCode }` : languageCode;
+		const iframeLanguage = () => {
+			const { iframe: { language } = {} } = this.props;
+			return language;
 		};
 
-		I18n.changeLocale(normalizeLanguageString(configLanguage() || browserLanguage()));
+		I18n.changeLocale(Language.normalizeLanguageString(iframeLanguage() ? iframeLanguage() : configLanguage() || Language.browserLanguage()));
 		I18n.on('change', this.handleLanguageChange);
 
 		Livechat.onStreamData('connected', this.handleConnected);
@@ -193,6 +182,7 @@ export class App extends Component {
 
 	async finalize() {
 		CustomFields.reset();
+		Language.reset();
 		userPresence.reset();
 		visibility.removeListener(this.handleVisibilityChange);
 		I18n.off('change', this.handleLanguageChange);
@@ -213,6 +203,7 @@ export class App extends Component {
 		expanded,
 		alerts,
 		modal,
+		iframe,
 	}, { initialized }) => {
 		if (!initialized) {
 			return null;
@@ -228,6 +219,7 @@ export class App extends Component {
 			sound,
 			alerts,
 			modal,
+			iframe,
 			onEnableNotifications: this.handleEnableNotifications,
 			onDisableNotifications: this.handleDisableNotifications,
 			onMinimize: this.handleMinimize,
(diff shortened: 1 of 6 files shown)


src/components/App/index.js
```diff
@@ -143,6 +144,7 @@ export class App extends Component {
 		await loadConfig();
 		this.handleTriggers();
 		CustomFields.init();
+		Language.init();
```
renatobecker: Why did you create a `class` to handle Language stuff instead of just exporting the methods you need to use?

src/lib/hooks.js
```diff
@@ -1,5 +1,6 @@
 import Triggers from './triggers';
 import CustomFields from './customFields';
+import Language from './language';
```
renatobecker: You don't need a `class` to call just one method, you can create a `file` and export the method there.

src/lib/hooks.js
```diff
@@ -108,6 +109,10 @@ const api = {
 
 		createOrUpdateGuest(data);
 	},
+  
+	setLanguage(language) {
+		Language.setLanguage(language);
```
renatobecker: Please, `export` the `setLanguage` method rather than instantiate a `class` to call the method.

src/lib/language.js
```diff
@@ -0,0 +1,68 @@
+import store from '../store';
+
+class Language {
```
renatobecker: You don't need all of this stuff. The idea is to export the method(s) you want to call.
Please fix it.

knrt10:
Updated changes

src/lib/language.js
```diff
@@ -0,0 +1,47 @@
...
+ */
+export const setLanguage = (language) => {
+	const { iframe } = store.state;
+	store.setState({ iframe: { ...iframe, language } });
+};
+
+/**
+ * To get browser Language of user
+ */
+export const browserLanguage = () => {
```
renatobecker: You don't need to export this method.

src/lib/language.js
```diff
@@ -0,0 +1,47 @@
...
+	navigator.userLanguage || navigator.language;
+};
+
+/**
+ * To normalize Language String and return language code
+ * @param {String} languageString
+ */
+export const normalizeLanguageString = (languageString) => {
```
renatobecker: You don't need to export this method.

src/components/App/index.js
```diff
@@ -163,25 +164,12 @@ export class App extends Component {
...
-			languageCode = languageCode.toLowerCase();
-
-			if (!countryCode || countryCode.length !== 2) {
-				countryCode = null;
-			} else {
-				countryCode = countryCode.toUpperCase();
-			}
-
-			return countryCode ? `${ languageCode }_${ countryCode }` : languageCode;
+		const iframeLanguage = () => {
```
renatobecker: Since you're calling `updateWidgetLanguage()`, I think you don't need this block of code here, right?

src/components/App/index.js
```diff
@@ -163,25 +164,12 @@ export class App extends Component {
...
+			const { iframe: { language } = {} } = this.props;
+			return language;
 		};
 
-		I18n.changeLocale(normalizeLanguageString(configLanguage() || browserLanguage()));
+		I18n.changeLocale(normalizeLanguageString(iframeLanguage() ? iframeLanguage() : configLanguage() || browserLanguage()));
```
renatobecker: Move this code into `updateWidgetLanguage` method.

knrt10: ok

src/components/App/index.js
```diff
@@ -210,6 +188,7 @@ export class App extends Component {
 		expanded,
 		alerts,
 		modal,
+		iframe,
```
renatobecker: Is this is necessary?

src/lib/hooks.js
```diff
@@ -108,6 +109,11 @@ const api = {
 
 		createOrUpdateGuest(data);
 	},
+  
+	setLanguage(language) {
+		setLanguage(language);
+		updateWidgetLanguage();
```
renatobecker: Call this method inside `setLanguage` method.

src/lib/language.js
```diff
@@ -0,0 +1,55 @@
...
+const configLanguage = () => {
+	const { config: { settings: { language } = {} } = {} } = store.state;
+	return language;
+};
+
+/**
+ * This will update langauge of widget
+ */
+export const updateWidgetLanguage = () => {
+	const { iframe: { language } = {} } = store.state;
```
renatobecker: Why not dealing with this information on `configLanguage` method?
You don't need to read the `store` so many times since you're reading the store on `configLanguage` method.
Please, test the `iframe` data on `configLanguage` method.

src/lib/language.js
```diff
@@ -0,0 +1,55 @@
...
+	I18n.changeLocale(language ? language : normalizeLanguageString(configLanguage() || browserLanguage()));
```
renatobecker:
```suggestion
	I18n.changeLocale(normalizeLanguageString(configLanguage() || browserLanguage()));
```

Related to the comment I left above..

src/lib/language.js
```diff
@@ -0,0 +1,55 @@
...
+ */
+const configLanguage = () => {
+	const { config: { settings: { language } = {} } = {} } = store.state;
+	return language;
+};
+
+/**
+ * This will update langauge of widget
+ */
+export const updateWidgetLanguage = () => {
```
renatobecker: Let's replace the name of this method by `setWidgetLanguage`.

src/lib/language.js
```diff
@@ -0,0 +1,54 @@
...
+		countryCode = countryCode.toUpperCase();
+	}
+
+	return countryCode ? `${ languageCode }_${ countryCode }` : languageCode;
+};
+
+/**
+ * To get browser Language of user
+ */
+const browserLanguage = () => {
```
renatobecker:
```suggestion
const browserLanguage = () => (navigator.userLanguage || navigator.language);
```

Use the correct arrow function statement. Your function does not have a return.

src/lib/hooks.js
```diff
@@ -108,6 +109,10 @@ const api = {
 
 		createOrUpdateGuest(data);
 	},
+  
+	setLanguage(language) {
```
renatobecker: You don't need this method here, you can set the iframe language on the `setLanguage` created on `hooks.js`.

Issue #196, opened 2019-03-15
Register Form does not disappear after calling registerGuest API method


renatobecker:
After calling the `registerGuest`(https://rocket.chat/docs/developer-guides/livechat-api/#register-visitor) API method, the Register Form needs to disappear and the Chat Screen needs to show up.

PR #216, opened 2019-04-18
fixed RegisterGuest Api bug #196

Jagansivam28:
added lifecycle method componentDidUpdate for checking if Guest Registration Update for showing chat page if Guest Registered API

### DIFF
--- a/src/components/App/index.js
+++ b/src/components/App/index.js
@@ -65,6 +65,7 @@ export class App extends Component {
 			if (showRegistrationForm) {
 				return route('/register');
 			}
+			return route('/');
 		}, 100);
 	}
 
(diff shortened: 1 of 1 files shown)


src/components/App/index.js
```diff
@@ -203,6 +203,12 @@ export class App extends Component {
 		this.finalize();
 	}
 
+	componentDidUpdate(previousProps, previousState, previousContext) {
```
knrt10: I think if 2nd and third params are not used, you should remove it.

renatobecker:
Fixed by https://github.com/RocketChat/Rocket.Chat.Livechat/pull/221.

PR #220, opened 2019-04-26
Support thread messages as attachments

renatobecker:
Before the implementation of threads, the payload of `reply` and `quoted` messages were equal, but now the reply messages are not sent as attachments anymore, so we need to fetch the parent message from the server and then render it as an attachment.

### DIFF
--- a/src/lib/room.js
+++ b/src/lib/room.js
@@ -122,7 +122,7 @@ export const loadMessages = async() => {
 	}
 
 	await store.setState({ loading: true });
-	const messages = await Livechat.loadMessages(rid).then(async(data) => (normalizeMessages(data)));
+	const messages = await Livechat.loadMessages(rid).then((data) => (normalizeMessages(data)));
 	await initRoom();
 	await store.setState({ messages: (messages || []).reverse(), noMoreMessages: false });
 	await store.setState({ loading: false });
@@ -141,7 +141,7 @@ export const loadMoreMessages = async() => {
 	}
 
 	await store.setState({ loading: true });
-	const moreMessages = await Livechat.loadMessages(rid, { limit: messages.length + 10 }).then(async(data) => (normalizeMessages(data)));
+	const moreMessages = await Livechat.loadMessages(rid, { limit: messages.length + 10 }).then((data) => (normalizeMessages(data)));
 	await store.setState({
 		messages: (moreMessages || []).reverse(),
 		noMoreMessages: messages.length + 10 > moreMessages.length,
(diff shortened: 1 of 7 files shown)


src/lib/room.js
```diff
@@ -117,7 +122,7 @@ export const loadMessages = async() => {
 	}
 
 	await store.setState({ loading: true });
-	const messages = await Livechat.loadMessages(rid);
+	const messages = await Livechat.loadMessages(rid).then(async(data) => (normalizeMessages(data)));
```
diegolmello: Unnecessary async

src/lib/room.js
```diff
@@ -136,7 +141,7 @@ export const loadMoreMessages = async() => {
 	}
 
 	await store.setState({ loading: true });
-	const moreMessages = await Livechat.loadMessages(rid, { limit: messages.length + 10 });
+	const moreMessages = await Livechat.loadMessages(rid, { limit: messages.length + 10 }).then(async(data) => (normalizeMessages(data)));
```
diegolmello: Unnecessary async

src/lib/threads.js
```diff
@@ -0,0 +1,68 @@
...
+	let parentMessage = messages.find((msg) => msg._id === message.tmid);
+	if (!parentMessage) {
+		parentMessage = await findParentMessage(message.tmid);
+	}
+	const { msg, attachments = [] } = parentMessage;
+	return Object.assign(message, { threadMsg: parentMessage, attachments: [{ attachments, text: msg, tmid: message.tmid }] });
+};
+
+export const normalizeMessage = async(message) => {
+	if (await isThreadMessage(message)) {
```
diegolmello: Feels weird to me.
Can you run `await isThreadMessage(message)` first and then test the result?

src/lib/threads.js
```diff
@@ -0,0 +1,68 @@
...
+		return false;
+	}
+
+	if (message && message.tmid && !message.threadMsg) {
+		return await normalizeThreadMessage(message);
```
diegolmello: Unnecessary await

src/lib/threads.js
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	return message;
+};
+
+export const normalizeMessages = async(messages = []) => await Promise.all(messages.filter(async(message) => normalizeMessage(message)));
```
diegolmello: The second async is unnecessary

renatobecker:
@diegolmello, fixed!
Thanks.

PR #221, opened 2019-04-26
Close registration form when registering the user through the Livechat API

### LINKED ISSUE #196: Register Form does not disappear after calling registerGuest API method

renatobecker:
Closes #196.

Since the Livechat Router is not based on store/state, when a user is registered through the Livechat API, the registration form needs to detect this change and redirect it to `Chat` container.

### DIFF
--- a/src/routes/Register/container.js
+++ b/src/routes/Register/container.js
@@ -8,10 +8,6 @@ import Register from './component';
 
 export class RegisterContainer extends Component {
 
-	state = {
-		registered: false,
-	}
-
 	getDepartment = (fields = {}) => {
 		let { department } = fields;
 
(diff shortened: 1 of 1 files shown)


src/routes/Register/container.js
```diff
@@ -46,6 +50,16 @@ export class RegisterContainer extends Component {
 		}
 	}
 
+	componentDidUpdate(prevProps) {
+		const { user: prevUser } = prevProps;
+		const { user } = this.props;
+		const { registered } = this.state;
+
+		if (!registered && !prevUser && user) {
```
diegolmello: Do we really need `this.state.registered`?
I mean, if the user is on Register view and doesn't have a `prevUser` yet, he's not registered.
Then it would be just a matter of testing `prevUser` against `user`.

Everything else LTGM.

renatobecker: @diegolmello I have made other improvements on the codebase, so yeah, we don't need the state anymore.
Thanks.

PR #226, opened 2019-04-30
[WIP] Emoji rendering and addition

knrt10:
This PR will try to add emoji feature to live chat

- [x] Render emoji 
- [x] Add custom emojis
- [ ] Add emoji picker frame


renatobecker:
@knrt10, there are some emojis not supported, like `:slight_smile:`.
Did you try it?

renatobecker:
@knrt10,

There is a new property returned on the `config` method providing a list of RC custom emojis:

```"resources": {
"sound": "http://localhost:3000/sounds/chime.mp3",
"emojis": [
    {
        "_id": "3HN5spmYnmiijj4Jq",
        "name": "slight_smile",
        "aliases": [
            "slight_smile;slight_aa"
         ],
        "extension": "png",
        "_updatedAt": "2019-05-03T13:12:55.322Z"
    }
],
},```

So, we'll need to deal with this kind of emoji as well and get it from the server when necessary.

File: src/components/Messages/MessageText/index.js [hunk 1]
```diff
@@ -11,7 +12,7 @@ export const MessageText = memo(({
 }) => (
 	<div
 		// eslint-disable-next-line react/no-danger
-		dangerouslySetInnerHTML={{ __html: renderMarkdown(text) }}
+		dangerouslySetInnerHTML={{ __html: emojify(renderMarkdown(text), { output: 'unicode' }) }}
```
renatobecker: Why not applying the `emojiify` stuff into `renderMarkdown` method?

knrt10: hmm, that would be better

renatobecker:
@knrt10, please follow the Tasso's instructions(last comment) and, once you ensure that everything is working fine, you need to add the message emoji rendering on Storybook as well.

knrt10:
@tassoevan I am sorry I did not get it. Did you mean to check compatibility with Preact I should check in Storybook?

tassoevan:
@knrt10 Let me clarify this: since Storybook supports only React, to use it we have to bridge our components via `preact-compat`. The final client bundle, however, doesn't have this compatibility layer, and things that are supposed to work over React only will work only on Storybook. You need to run the client, via `yarn start`, to check if the dependencies work over Preact alone.

knrt10:
Thanks @tassoevan got it. I ran `yarn start` it worked fine. But to be clear should I run in production i.e `NODE_ENV=production` inside `.env` file or just run as it is? I tried both ways it's working fine

src/components/Messages/MessageText/customEmoji.js
```diff
@@ -0,0 +1,30 @@
+import { store } from '../../../store';
+import { Livechat } from '../../../api';
+import render from 'preact-render-to-string';
+import { emojify } from 'react-emojione';
+
+function search(nameKey, myArray) {
```
renatobecker: Replace to arrow function.
`const search = (key, array = []) => {...`

src/components/Messages/MessageText/customEmoji.js
```diff
@@ -0,0 +1,30 @@
...
+	for (let i = 0; i < myArray.length; i++) {
```
renatobecker: Replace to `find` approach:
`return array.find((item) => item.name === key);`

renatobecker: You need to check the `aliases` property as well.

src/components/Messages/MessageText/customEmoji.js
```diff
@@ -0,0 +1,30 @@
...
+	}
+}
+
+export const customEmoji = (text) => {
+
+	const { emojis } = store.state;
+	text = emojify(text, { output: 'unicode' });
+	const message = text.split(/:/g);
+	const content = [];
+	for (let i = 0; i < message.length; i++) {
```
renatobecker: You can create an `arrow function` const and assign its return to `content` const.

src/components/Messages/MessageText/markdown.js
```diff
@@ -1,7 +1,8 @@
 import MarkdownIt from 'markdown-it';
-
+import { customEmoji } from './customEmoji';
```
renatobecker: Let's replace the `customEmoji` name by `customEmojify`.

knrt10:
Current status of picker.

Emoji picker

tassoevan:
@sampaiodiego Any remarks about EmojiOne here?

renatobecker:
Livechat already supports emoji rendering, so I'm closing this PR.

PR #242, opened 2019-06-08
setting location Data for user

knrt10:
@renatobecker please review


package.json
```diff
@@ -64,6 +64,7 @@
 	"dependencies": {
 		"@kossnocorp/desvg": "^0.2.0",
 		"@rocket.chat/sdk": "^1.0.0-alpha.28",
+		"axios": "^0.18.0",
```
tassoevan: There is no need to use `axios` since we are polyfilling the `fetch()` API.

src/components/App/index.js
```diff
@@ -155,11 +156,32 @@ export class App extends Component {
...
+		const location = await axios.get(`${ 'https://cors-anywhere.herokuapp.com/' }https://geoip-db.com/json/${ ip }`)
+			.then((res) => res.data)
+			.catch((err) => err);
+
+		const token = getToken();
+		return {
+			location,
+			token,
+		};
+	}
```
tassoevan: I've some suggestions:

1. Check if GDPR protection is enabled and the user has approved its data collection;
2. Use the own browser's geolocation API before using an external service;
3. If the geolocation API fails, request to `geoip-db.com` as a fallback;
4. Verify the limits for commercial use off `geoip-db.com` -- despite Rocket.Chat and this client being MIT; 
5. The `fetch()` API has some options to properly do CORS requests;
6. The ipify functionally can be implemented on Rocket.Chat's Livechat API, decreasing the number of requests needed.

knrt10: Thanks for the review, will update it accordingly

reetp:
Yes - this must not be automatically implemented, must have an off switch set to disabled by default, and be separately switchable from other information 

eg you may wish to collect an email address but NOT geo location data

There must be some way to notify users of the data collection if it is enabled.

src/components/App/index.js
```diff
@@ -155,11 +157,75 @@ export class App extends Component {
 		I18n.on('change', this.handleLanguageChange);
 	}
 
+	locationBackup = async() => {
```
renatobecker: Let's create a specific file/helper to provide this method.

src/components/App/index.js
```diff
@@ -155,11 +157,75 @@ export class App extends Component {
...
+		const { ip } = await fetch(`${ 'https://cors-anywhere.herokuapp.com/' }https://api.ipify.org?format=json`, {
+			mode: 'cors',
+			headers: {
+				'Access-Control-Allow-Origin': '*',
+			},
+		}).then((res) => (res.json()));
+
+		const location = await fetch(`${ 'https://cors-anywhere.herokuapp.com/' }https://geoip-db.com/json/${ ip }`, {
```
renatobecker: We need to define the service which will provide the geolocation data on the server side.
Hardcoding is not a good approach.

src/components/App/index.js
```diff
@@ -155,11 +157,75 @@ export class App extends Component {
...
+		}).then((res) => (res.json()));
+	
+		const token = getToken();
+		return {
+			location,
+			token,
+		};
+	}
+	
+	locationPrimary = async(latitude, longitude) => {
```
renatobecker: Move this code to a specific file/helper.

src/components/App/index.js
```diff
@@ -155,11 +157,75 @@ export class App extends Component {
...
+			token,
+		};
+	}
+
 	async initialize() {
 		// TODO: split these behaviors into composable components
 		// Call loadConfig before calling Livechat.connect
 		await loadConfig();
 		await Livechat.connect();
+		const checkLocationUser = await Livechat.checkLocationUser(getToken());
```
renatobecker: Move this code to a specific file/helper.

src/widget.js
```diff
@@ -311,7 +311,7 @@ const init = (url) => {
 	}
 
 	config.url = url;
-
+
```
renatobecker: Remove the unnecessary blank space.

src/components/App/index.js
knrt10: ok

src/components/App/index.js
knrt10: Ok, let me do that

src/components/App/index.js
knrt10: ok

src/lib/location.js
```diff
@@ -0,0 +1,194 @@
...
+	});
+
+/**
+ * This is used to get location details for user
+ * @param {Number} latitude
+ * @param {Number} longitude
+ * @returns {Object}
+ */
+const locationPrimary = async(latitude, longitude) => {
+	const { address } = await fetch(`https://nominatim.openstreetmap.org/reverse?format=json&lat=${ latitude }&lon=${ longitude }`, {
```
geekgonecrazy: So this is going out to this api to grab location?  This has to definitely be optional if done at all.  Some serious privacy concerns around sending data like this out.

src/lib/location.js
```diff
@@ -0,0 +1,194 @@
...
+		deviceInfo: deviceInfo(),
+	};
+};
+
+/**
+ * This is backup method to get location of user
+ * @returns {Object}
+ */
+const locationBackup = async() => {
+	const location = await fetch('https://api.ipdata.co?api-key=test', {
```
geekgonecrazy: Same here

reetp:
Apart from truck loads of GDPR concerns here this, wording has a typo.

> "please_enable_your_location_to_start_chat_this_is__c68e0f77": "Please enable your location to start chat. This is for your convinience only. Thanks",

"Please enable your location to start chat. This is for your **convenience** only. Thanks"

It also sounds like you are telling users that unless they enable location they cannot chat with you?

I don't think so.

- You have to ask for permission before you take any data. You can't do it in the background and then say "oh please let me use this"

- If you want to insist on obtaining location data before chatting then it needs an off switch for those that do not want to impose this

- You have to be able to remove their data on request if you are storing it

src/api.js
```diff
@@ -1,9 +1,11 @@
 import queryString from 'query-string';
-import LivechatClient from '@rocket.chat/sdk/lib/clients/Livechat';
 
-const host =	window.SERVER_URL
-	|| queryString.parse(window.location.search).serverUrl
-	|| (process.env.NODE_ENV === 'development' ? 'http://localhost:3000' : null);
+// eslint-disable-next-line import/no-absolute-path
+import LivechatClient from '/Users/knrt10/dev/openSource/RocketChat/Rocket.Chat.js.SDK/dist/lib/clients/Livechat';
```
renatobecker: I think you forget to remove this code before pushing your changes.

src/routes/Chat/container.js
```diff
@@ -106,13 +106,15 @@ export class ChatContainer extends Component {
 			this.stopTypingDebounced.stop();
 			await Promise.all([
 				this.stopTyping({ rid, username: user.username }),
+				Livechat.updateSessionStatus('online', token),
```
renatobecker: Why call this method here?
We have a specific lib to deal with user status.

src/routes/Register/container.js
```diff
@@ -30,6 +30,7 @@ export class RegisterContainer extends Component {
 		await dispatch({ loading: true, department });
 		try {
 			await Livechat.grantVisitor({ visitor: { ...fields, token } });
+			await Livechat.updateVisitorSessionOnRegister({ visitor: { ...fields, token } });
```
renatobecker: Why not just one method to update the livechat session? 
You just need one method, you can pass specific data as parameters.

src/components/App/index.js
```diff
@@ -149,6 +151,7 @@ export class App extends Component {
 	async finalize() {
 		CustomFields.reset();
 		userPresence.reset();
+		userSessionPresence.reset();
```
renatobecker: We don't need this lib.

renatobecker:
@knrt10 
There is a lot of work to do here.

- We just need one new method on the SDK, I suggest to call it as `Livechat.updateVisitorSession`.
- Regardless you send the user location or not, we can pass all of the information through the SDK method parameter;
- You need to create a new setting on the backend side which will be used to allow/deny ask the user to collect its location on the widget side;

Please, keep in mind you need to reduce the complexity in this PR, one suggestion is to use just one method to handle the session data.
Feel free to ping me in case any of my comments are not clear for you.

engelgabriel:
https://github.com/RocketChat/feature-requests/issues/135

PR #324, opened 2019-11-07
Update polish translation

vizarch:


maciej87pl:
when it will be added to livechat widget? In latest version is old translation

PR #348, opened 2019-12-19
[NEW] Translated to Hebrew

idocme:
This language is RTL! so when using this. the livechat should be right to left.


ggazzo:
@idocme there are some lines breaking the json file... please check on the develop brach,I restored the default language and I inserted a small script to change the text direction

PR #359, opened 2020-01-23
[NEW] Russian translation

v1adimir:
Almost all string translated to Russian.


m0rg0t:
As i can see at current moment Russian localization isn't merged with develop branch?

PR #363, opened 2020-02-12
Corrected German title for  finished chat

ruKurz:
This pull request corrects the German upper and lower case for the label of finished chat.


ruKurz:
I executed `yarn run i18n` but this does change any file to be commited.

PR #370, opened 2020-02-14
[FIX] Spanish translation

imiguelsalcedo:
Full translation to _Spanish_ language.


ggazzo:
@imiguelsalcedo hey thanks for you contribution, could you check please and resolve the conflicts, I'm afraid we merged some ES dictionary already but we can improve that one :)

MartinSchoeler (approved):
Looks good to me now!

PR #380, opened 2020-03-09
[FIX] Livechat guest avatar using name instead of username

### LINKED ISSUE #369: Wrong avatar displayed (apart from on left sidebar in web app)
I guess this may be related to #237 (and potentially to #368 too) but it's a different issue.

I was just testing my livechat set-up using name 'Testing livechat' but the avatar showing in the chat window on the website was a G instead of a T:

Wrong_avatar

Confusingly when looking at the web client as the agent it was still a G in the main window, but correctly shows as a T in the sidebar:

wrong_avatar_2

Perhaps something to do with the avatar of the closed but still stuck in the left sidebar livechat conversation (as per #368 ) being a chat with a person with a G avatar?

ashwaniYDV:
closes #369

### DIFF
--- a/src/components/Messages/Message/index.js
+++ b/src/components/Messages/Message/index.js
@@ -68,14 +68,17 @@ const getSystemMessageText = ({ t, conversationFinishedMessage }) =>
 	|| (t === MESSAGE_TYPE_LIVECHAT_CLOSED && (conversationFinishedMessage || I18n.t('Conversation finished')));
 
 const getMessageUsernames = (compact, message) => {
-	if (compact) {
+	if (compact || !message.u) {
 		return [];
 	}
-	if (message.alias && message.u && message.u.name) {
-		return [message.u.name];
+
+	const { alias, u: { username, name } } = message;
+	if (alias && name) {
+		return [name];
 	}
-	return message.u && message.u.username && [message.u.username];
-}
+
+	return [username];
+};
 
 export const Message = memo(({
 	avatarResolver,
--- a/src/components/Messages/Message/index.js
+++ b/src/components/Messages/Message/index.js
@@ -67,6 +67,16 @@ const getSystemMessageText = ({ t, conversationFinishedMessage }) =>
 	|| (t === MESSAGE_TYPE_WELCOME && I18n.t('Welcome'))
 	|| (t === MESSAGE_TYPE_LIVECHAT_CLOSED && (conversationFinishedMessage || I18n.t('Conversation finished')));
 
+const getMessageUsernames = (compact, message) => {
+	if (compact) {
+		return [];
+	}
+	if (message.alias && message.u && message.u.name) {
+		return [message.u.name];
+	}
+	return message.u && message.u.username && [message.u.username];
+}
+
 export const Message = memo(({
 	avatarResolver,
 	attachmentResolver = getAttachmentUrl,
@@ -88,8 +98,7 @@ export const Message = memo(({
 	>
 		<MessageAvatars
 			avatarResolver={avatarResolver}
-			// eslint-disable-next-line no-nested-ternary
-			usernames={compact ? [] : message.alias && message.u && message.u.name ? message.u && message.u.name && [message.u.name] : message.u && message.u.username && [message.u.username]}
+			usernames={getMessageUsernames(compact, message)}
 		/>
 		<MessageContent reverse={me}>
 			{renderContent({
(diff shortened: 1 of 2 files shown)


ashwaniYDV:
@renatobecker 
Do I need to place this `// eslint-disable-next-line no-nested-ternary` in that particular line only or include it in the global settings

ashwaniYDV:
Also do we need to do a check for `name` property too? It is a required filed for a guest to establish communication with agent, so it will always be there. Please suggest

src/components/Messages/Message/index.js
```diff
@@ -88,7 +88,8 @@ export const Message = memo(({
 	>
 		<MessageAvatars
 			avatarResolver={avatarResolver}
-			usernames={compact ? [] : message.u && message.u.username && [message.u.username]}
+			// eslint-disable-next-line no-nested-ternary
+			usernames={compact ? [] : message.alias ? message.u && message.u.name && [message.u.name] : message.u && message.u.username && [message.u.username]}
```
renatobecker: I think extending ternary operators is making hard to read the logic. My suggestion is to create a helper method to deal and return the expected `usernames`.

```suggestion
			usernames={this.getMessageUsernames(message)}
```

renatobecker:
> Also do we need to do a check for `name` property too? It is a required filed for a guest to establish communication with agent, so it will always be there. Please suggest

No, the `name` property may not be available, it's possible to disable the registration form.

ashwaniYDV:
> Can you please add a new story to `storybook` representing this use case?
> Thanks.

Sorry @renatobecker I didn't get it. Where should I add the story. I have done the other requested changes.

src/components/Messages/Message/index.js
```diff
@@ -67,6 +67,16 @@ const getSystemMessageText = ({ t, conversationFinishedMessage }) =>
 	|| (t === MESSAGE_TYPE_WELCOME && I18n.t('Welcome'))
 	|| (t === MESSAGE_TYPE_LIVECHAT_CLOSED && (conversationFinishedMessage || I18n.t('Conversation finished')));
 
+const getMessageUsernames = (compact, message) => {
+	if (compact) {
+		return [];
+	}
+	if (message.alias && message.u && message.u.name) {
```
renatobecker: Why do check the `alias` property?

renatobecker:
> > Can you please add a new story to `storybook` representing this use case?
> > Thanks.
> 
> Sorry @renatobecker I didn't get it. Where should I add the story? I have done the other requested changes.

Here -> https://github.com/RocketChat/Rocket.Chat.Livechat/blob/develop/src/components/Messages/MessageAvatars/stories.js

src/components/Messages/Message/index.js
ashwaniYDV: because alias is there only for guest and not for agent

ashwaniYDV:
> > Also do we need to do a check for `name` property too? It is a required filed for a guest to establish communication with agent, so it will always be there. Please suggest
> 
> No, the `name` property may not be available, it's possible to disable the registration form.

So if name property is not there then alias property will also be not there (I guess) and default username will be the avatar.

src/components/Messages/Message/index.js
ashwaniYDV: Here alias is the possible way to distinguish between guest and agent

renatobecker: No. If you check the server-side, the alias is available in other situations.
My suggestion is not to check it.

ashwaniYDV: Then how can we distinguish between the guest and agent since name and username properties are common to both. 
Should I check it with regex of `guest-xx` or there is some other way which I don't know? Please help

renatobecker: Okay. I'll suggest another implementation.

src/components/Messages/Message/index.js
```diff
@@ -67,6 +67,16 @@ const getSystemMessageText = ({ t, conversationFinishedMessage }) =>
 	|| (t === MESSAGE_TYPE_WELCOME && I18n.t('Welcome'))
 	|| (t === MESSAGE_TYPE_LIVECHAT_CLOSED && (conversationFinishedMessage || I18n.t('Conversation finished')));
 
+const getMessageUsernames = (compact, message) => {
```
renatobecker:
```suggestion
const getMessageUsernames = (compact, message) => {
	if (compact || !message.u) {
		return [];
	}
	
	const { alias, u: { username, name} } = message;
	if (alias && name) {
        	return name;
	}
	
	return username;

```

ashwaniYDV:
I have done the chnages @renatobecker. Please review. Also have I added the story corrected?

PR #382, opened 2020-03-13
[FIX] On mobile Livechat covers the whole screen, making it intrusive for triggers

### LINKED ISSUE #244: On mobile Livechat covers the whole screen, making it intrusive for triggers.
If I setup a trigger for the livechat, say, with people that stayed 10 sec on the webpage, the chat window appears. Thought on desktop the window is not that obstructive, on mobile it is made full screen.

I think there should be an option to trigger popped down, such that the user can actually "accept" the interaction. Otherwise it is very intrusive.

image

Like this?

image

ashwaniYDV:
closes #244 #386

#### Description:
I've added a 5sec timeout trigger.


src/lib/triggers.js
```diff
@@ -129,7 +137,10 @@ class Triggers {
 
 					route('/');
 					parentCall('openWidget');
-					store.setState({ minimized: false });
+
+					if (!window.mobilecheck()) {
```
renatobecker: Instead of defining `minimized: false`, wouldn't it be better to skip calling the `parentCall` method?

ashwaniYDV: That would be better (y)

But we have to skip both parentcall and setstate({minimized: false]) to not open widget in mobiles.

renatobecker: By `skipping` I meant `return`...

src/lib/triggers.js
```diff
@@ -9,6 +9,14 @@ import { normalizeAgent } from './api';
 
 const agentCacheExpiry = 3600000;
 let agentPromise;
+
+window.mobilecheck = function() {
```
renatobecker: Please, move this code/function to the `helper` file.

ashwaniYDV: By helpers did you mean helpers.stories.js?
Or should I create a new file `scr/helpers.js` as there is no helpers.js file

ashwaniYDV:
@renatobecker Please review

src/lib/triggers.js
ashwaniYDV: done (y)

dlecan:
It’s an interesting feature. Is it possible finish/merge it?

dlecan:
@ashwaniYDV :+1: 

@renatobecker?

patrick-tb:
Is there any news in this regard?

@ashwaniYDV 
@renatobecker

renatobecker:
Fixed by -> https://github.com/RocketChat/Rocket.Chat.Livechat/pull/553

PR #383, opened 2020-03-17
[NEW] Add new API method the set the default Agent before chatting

renatobecker:
Adds the `setAgent` method which allows setting the default agent before the visitor starts chatting.
Also, since the agent who sends trigger messages is the same who will serve the conversation, the store prop `triggerAgent` has been renamed/unified to `defaultAgent` prop.

### DIFF
--- a/src/widget.js
+++ b/src/widget.js
@@ -269,7 +269,10 @@ function initialize(params) {
 			case 'language':
 				setLanguage(params[method]);
 				continue;
-			default:
+				case 'agent':
+					setAgent(params[method]);
+					continue;
+				default:
 				continue;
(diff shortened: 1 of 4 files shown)


src/widget.js
```diff
@@ -269,7 +269,10 @@ function initialize(params) {
 			case 'language':
 				setLanguage(params[method]);
 				continue;
-			default:
+				case 'agent':
+					setAgent(params[method]);
+					continue;
+				default:
```
tassoevan:
```suggestion
			case 'agent':
				setAgent(params[method]);
				continue;
			default:
```

PR #399, opened 2020-04-24
Improvements on sending offline messages

### LINKED ISSUE #396: More prominent dialog upon successful offline form submission
To prevent duplicate form submissions and people who might not see the notification that fades in from the top when they send the form, it would be nice to maybe show an dialog instead:

Screen Shot 2020-04-20 at 12 00 49

Or hide the form elements completely when the message was sent successfully.

renatobecker:
CLOSES #396 

Improvements on the Offline Message Form:
- Show a modal message after the message is sent successfully instead of an alert on the top of the form. The previous alert was not clear enough and users where clicking on the submit button several times. Now the users will get a modal alert like shown below:

Screen Shot 2020-04-24 at 17 48 42

- We're now resetting the form after the message is sent;
- We're now sharing the `hostname` from where the message has been sent;

### DIFF
--- a/src/components/helpers.js
+++ b/src/components/helpers.js
@@ -126,7 +126,7 @@ export const sortArrayByColumn = (array, column, inverted) => array.sort((a, b) 
 });
 
 export const parseOfflineMessage = (fields = {}) => {
-	const host = location.protocol.concat('//').concat(window.location.hostname);
+	const host = window.location.origin;
 	return Object.assign(fields, { host });
 };
 export const normalizeDOMRect = ({ left, top, right, bottom }) => ({ left, top, right, bottom });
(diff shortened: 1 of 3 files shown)


src/components/helpers.js
```diff
@@ -125,6 +125,10 @@ export const sortArrayByColumn = (array, column, inverted) => array.sort((a, b)
 	return 1;
 });
 
+export const parseOfflineMessage = (fields = {}) => {
+	const host = location.protocol.concat('//').concat(window.location.hostname);
```
tassoevan, line 129:
```suggestion
	const host = window.location.origin;
```

vladislavs:
This is really crucial issue. We've been suffering because it, clients spaming our helpdesk with same input values. Success message is not enough visible and text inputs should be cleaned after sent. Please force the merge into master.

PR #408, opened 2020-05-09
[NEW] adding support for RichMessages 

murtaza98:
The main goal is to add support for RichMessages in Livechat.

Demo for the results so far
image

Tasks:-
- [x] Create a new RichMessageGroup Component
- [x] Link RichMessageGroup with rest of the components
- [ ] Add onClick Listeners to RichMessage Buttons
- [ ] Link RichMessageGroup component with LiveChat API

### DIFF
--- a/src/components/RichMessageGroup/index.js
+++ b/src/components/RichMessageGroup/index.js
@@ -25,11 +25,13 @@ export class RichMessageGroup extends MemoizedComponent {
 	render = ({
 		className,
 		style = {},
+		richMessages,
 	}) => (
 		<div
 			className={createClassName(styles, 'rich-message-group', {}, className)}
 			style={style}
 		>
+			{console.log("From richMessageGroup "+richMessages)}
 			{this.renderItems(this.props).map((child) => cloneElement(child, { className: createClassName(styles, 'rich-message-group__item') }))}
 		</div>
 	)
--- a/src/components/RichMessageGroup/index.js
+++ b/src/components/RichMessageGroup/index.js
@@ -1,0 +1,36 @@
+import { cloneElement } from 'preact';
+
+import { createClassName, memo, MemoizedComponent } from '../helpers';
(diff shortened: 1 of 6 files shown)


renatobecker:
Hi @murtaza98 I'm really glad to know that you're willing to work on `Rich Messages` stuff.
The fact is we already have a prototype for  the `Rich Messages` design, as you can see below:

Screen Shot 2020-05-09 at 11 23 11
Screen Shot 2020-05-09 at 11 22 59
Screen Shot 2020-05-09 at 11 22 53
Screen Shot 2020-05-09 at 11 22 47

So, to be able to accept this implementation we'll need to follow the design system created for this feature.
I can share it with you if you're interested to implement the `Rich Messages` components according to the the design system expected.

murtaza98:
Hey, @renatobecker Thanks for the feedback.
Yes, it would be helpful, if u share the expected design system. I will try to implement it :)

murtaza98:
Also a quick question, 
I found [this](https://github.com/WideChat/Rocket.Chat.Android/wiki/Phase-1--Rich-Messaging-Acceptance-Test) article ... it's of RocketChat's Android app.

It says that the rich messages content, which we receive from RC server, it will be a part of **attachments** -> **actions** property of message object, something like this
```
message Object
{
  "alias": "Gruggy",
  "roomId": "Xnb2kLD2Pnhdwe3RH",
  "text": "Sample message",
  "attachments": [
    {
      "actions": [
        {
          "type": "button",
          "text": "Rich Message Option 1",
          "url": "http://www.kayak.com",
          "is_webview": false
        }
        {
          "type": "button",
          "text": "Rich Message Option 2",
          "url": "http://www.kayak.com",
          "is_webview": false
        }
      ]
    }
  ]
}
```

So is that correct?

ggazzo:
check ui-kit on rocket.chat

PR #412, opened 2020-05-13
 NEW] Add Emoji rendering support

### LINKED ISSUE #335: Livechat Widget does not display markdown correctly
### Description:

When a live chat agent uses markdown in order to structure and beautify its replys, the customer who uses the livechat widget will see the markdown syntax. 

### Actual behavior:

Examples:

- Bold text will be displayed with stars (*) at the widget side
- Italic text will be displayed with underscores (_) at the widget side
- ....

The biggest problem with this issue is, that the agent thinks that the client has the same representation of the text.

This issue is closly related to https://github.com/RocketChat/Rocket.Chat/issues/7700

### Expected behavior:

Full support of markdown in the widget.

### Server Setup Information:

* Version of Rocket.Chat Server: 0.56.0
* Operating System: Linux
* Deployment Method(snap/docker/tar/etc): AWS
* Number of Running Instances: 2
* Node Version: v4.8.2

### Steps to Reproduce:
Format text with markdown in livechat. The widget will not display it correctly.

Additional information:
I think there is an issue which has the same scope: https://github.com/RocketChat/Rocket.Chat/issues/3796 However this one is not described very well.

### LINKED ISSUE #409: [Question] Does messages in Livechat Support have support for emojis?
Hello,
While working with Livechat feature, I noticed that the LiveAgent who has access to Rocket.Chat server chat interface can send an emojis to a Livechat Visitor.  However, the Livechat Visitor who is using the LiveChat Chat Interface is not able to properly render the emojis. Instead, it was showing only the placeholder title of emojis instead of actual emojis.
Have a look at SS

This is interface of LiveAgent
image

And this is the interface of LiveChat Visitor
image

As u can see, Instead of showing the emoji icon :smile: it is displaying this ```:smile:```

I think that this problem is bcs the Livechat message component doesn't support emojis. Am I correct? 
Also, I am planning to work on adding emoji support to Livechat, so if there is any resource or suggestions u would like to share, I would appreciate it :raised_hands:

Thanks :+1:

### LINKED ISSUE #445: Live Chat: guests cannot see emojis
### Description:

When an operator uses a emoji in live chat, the guest see's only the text, and not the image. 

### Server Setup Information:

* Version of Rocket.Chat Server: 0.58.2
* Operating System: Ubuntu 16.04 LTS
* Deployment Method(snap/docker/tar/etc): snap
* Number of Running Instances:  1
* DB Replicaset Oplog: ?
* Node Version: 4.8.3

### Steps to Reproduce:

1. Open new live chat between a guest and an operator.
2. As the operator, send a emoji in the chat

### Expected behavior:

Guest sees emoji as it should be displayed

### Actual behavior:

Guest see's text equivalent

### Screenshots
guest live chat

murtaza98:
CLOSES #335 
CLOSES #409 
CLOSES #445

The goal of this PR is to add support for Emoji to Livechat Widget.

The Tasks to be carried out include

- [x] add emoji Picker
- [x] link emoji Picker and Input text field
- [x] make the emoji Picker responsive
- [x] add support for rendering emoji to existing Message Component of Livechat #409


murtaza98:
Hello, I quick update
I have tried to make the emoji Picker responsive
I tested it on using chrome developer option on various phones, have a look

- my laptop
image


- IPhone X 
image

- iPhone 6/7/8 
image

- Pixel 2
image

- IPad
image

renatobecker:
Amazing job, @murtaza98 !
Congrats! 🥇

src/components/Composer/index.js
```diff
@@ -172,6 +173,16 @@ export class Composer extends Component {
...
 	}
 
+	componentWillMount() {
+		if(this.props.notifyEmojiSelect){
+			this.props.notifyEmojiSelect(this.handleNotifyEmojiSelect);
+		}
+	}
+
+	handleNotifyEmojiSelect(emojiColonName) {
+		this.el.innerText = this.el.innerText + emojiColonName;
```
renatobecker, line 183:
```suggestion
		this.el.innerText = `${ this.el.innerText } emojiColonName`;
```

murtaza98: Incorporated this change in lastest commit :+1: thanks for the feedback :smiley:

murtaza98:
Hi @renatobecker 
I'm glad to say that the above-mentioned tasks are all complete :rocket: :100: 

Have a look :smiley:  
On my laptop
image
On iPhone X
image

I hope u like this and any suggestions are most welcome :smiley:

renatobecker:
> Hi @renatobecker
> I'm glad to say that the above-mentioned tasks are all complete 🚀 💯
> 
> Have a look 😃
> On my laptop
> image
> On iPhone X
> image
> 
> I hope u like this and any suggestions are most welcome 😃

Hey, @murtaza98 this is an amazing job! 
I haven't done a deep review yet, I'm going to do that but I'd like to mention some UI/UX improvements that I noticed running your PR, such as:
- When the user clicks/selects an emoji on the emoji picker it should be closed automatically;
- When the user clicks/selects an emoji on the emoji picker and the selected emoji appears on the composer(text box), it should add an additional space after the emoji shortcut, like this:

Screen Shot 2020-05-15 at 09 29 40

Can you please apply those changes before review/merge the code?
Thanks and congratulations again!

murtaza98:
Hello @renatobecker 
There is a Setting for Picker on which I was hoping to get ur opinion.
So the Setting is named `SheetSize` and it defines the resolution of Picker. There are 4 options i.e 16, 20, 32, 64. As u might have guessed, Higher the resolution then the icon file size will be bigger and correspondingly, it will take more time to load on slow internet connections. 

I'd like to hear ur thoughts on this.
Thanks :smile:

murtaza98:
Hi @renatobecker
> * When the user clicks/selects an emoji on the emoji picker it should be closed automatically;
> * When the user clicks/selects an emoji on the emoji picker and the selected emoji appears on the composer(text box), it should add additional space after the emoji shortcut, like this:

Added both these changes in the latest commit :v:.

renatobecker:
> Hi @renatobecker
> 
> > * When the user clicks/selects an emoji on the emoji picker it should be closed automatically;
> > * When the user clicks/selects an emoji on the emoji picker and the selected emoji appears on the composer(text box), it should add additional space after the emoji shortcut, like this:
> 
> Added both these changes in the latest commit ✌️.

Cool! One more thing to improve:
When the user clicks outside the emoji picker, it should close automatically as well as when the user selects an emoji in the list. Did you get it?

Important question: It would be better if we don't use a third-party library to render the emoji, I mean, we should implement a native emoji render process. WDYT? 

Thanks,

murtaza98:
> Important question: It would be better if we don't use a third-party library to render the emoji, I mean, we should implement a native emoji render process. WDYT?

I'm sorry, but I don't get the benefit of this. I might be missing something...Could u share the concern?

Is ur concern regarding adding an additional module just for rendering?
If so then I'd like to mention that I've reused the same module which I've used for the Picker. So the Picker and rendering are both backed by only one module which is [emoji-mart](https://github.com/missive/emoji-mart)

renatobecker:
> > Important question: It would be better if we don't use a third-party library to render the emoji, I mean, we should implement a native emoji render process. WDYT?
> 
> I'm sorry, but I don't get the benefit of this. I might be missing something...Could u share the concern?
> 
> Is ur concern regarding adding an additional module just for rendering?
> If so then I'd like to mention that I've reused the same module which I've used for the Picker. So the Picker and rendering are both backed by only one module which is [emoji-mart](https://github.com/missive/emoji-mart)

Do you know what's this library size?
I talking about it because we have been facing some performance issues related to the `emoji one` library. I know we're not using `emoji one` here in this implementation, but we need to make sure the `emoji-mart`won't bring any performance issue to the widget. Okay?

murtaza98:
> Do you know what's this library size?

I checked their site, but couldn't find any official number. Then I checked the size of the package in my local node_modules folder and it is about 3.2 MB. Is that what u were asking?

> but we need to make sure the emoji-mart won't bring any performance issue to the widget. Okay?

Cool! Also since we are on this topic, I think we can tune the performance by reducing the pixel size of the rendered emojis. I'd mentioned more about it over [here](https://github.com/RocketChat/Rocket.Chat.Livechat/pull/412#issuecomment-629748319)

:v:

murtaza98:
Hi @renatobecker 

Added these 2 minor upgrades

1. Auto hider emoji picker when the user clicks outside the picker
2. Make emojis large while rendering when only emoji's are present in the text
image

renatobecker:
> Hi @renatobecker
> 
> Added these 2 minor upgrades
> 
> 1. Auto hider emoji picker when the user clicks outside the picker
> 2. Make emojis large while rendering when only emoji's are present in the text
>    image

Awesome! 

But I want to report this case that I just faced:

**Livechat:**
Screen Shot 2020-05-22 at 09 49 58

**Rocket.Chat:**
Screen Shot 2020-05-22 at 09 50 03

As you can see, there is a difference in the `emoji codes`, the Livechat sends `:woman-biking:` and Rocket.Chat sends `:woman_biking:`, we need to make both sides compatible.
Can you please take a look into it?

Thanks.

murtaza98:
Hi @renatobecker 
Regarding this

> But I want to report this case that I just faced:
> 
> **Livechat:**
> Screen Shot 2020-05-22 at 09 49 58
> 
> **Rocket.Chat:**
> Screen Shot 2020-05-22 at 09 50 03
> 
> As you can see, there is a difference in the `emoji codes`, the Livechat sends `:woman-biking:` and Rocket.Chat sends `:woman_biking:`, we need to make both sides compatible.
> Can you please take a look into it?

I looked into this, I found out that there is a few inconsistency between shortnames in `emojione` and `emoji-mart` package. In emoji-mart I found [this issue](https://github.com/missive/emoji-mart/issues/194)

murtaza98:
Hello,
Regarding [this](https://github.com/RocketChat/Rocket.Chat.Livechat/pull/412#issuecomment-633264164) issue of different short-names, I think I found a way using which we can solve it with minimal changes to this PR. If u feel this isn't a good design, it's totally fine. I too like the approach of using `unicodes` however, some blogs I read said that the `unicode` implementation varies from browser to browser. (I too experienced this, on my laptop, emojis r rendered black-white on `chrome`, whereas on `firefox` the same emojis appear colorful).

The above problem led me to think that since the `emoji-mart` package uses `svg-formatted-images` for rendering emojis, it would be an ideal solution that would work on all browsers.

I've also found a way in which we can solve the different short-names issue. The solution is a 2 step process:-

1. **Remove Conflicting Emojis From Picker**:- First, we should remove the conflicting named emojis from the `Picker`. We can easily maintain a list of all emojis whose short names r different on `emoji-one` and `emoji-mart` (I think this would be a shortlist). Examples of conflicting names are `man-bowing`, `woman-biking`, etc. So we will remove these from the Picker.

2.  **Replace emojis removed with new Custom Emojis with proper shortnames**:- In place of the removed emojis, we can add the same emojis with the correct name as it is on server to the `Picker`. In above eg, we will add new custom-emojis with name `man_bowing` and `women_biking` to the Picker. 

SideNote: The `emoji-mart` Picker allows us to perform both the above functions, I.e it allows us to add `custom-emojis` and also remove some pre-defined emojis

PR #425, opened 2020-06-05
[IMPROVE] Flow of the widget registration form

renatobecker:
When an automatic message is triggered, the Widget routes the visitor to the Chat Component, but when this happens and the end visitor starts typing it is registered as a guest because the registration form doesn't appear again. 

Current version:

Screen Shot 2020-06-05 at 20 10 11

To fix the issue we won't display the composer component in the `Chat Component` if the visitor is not registered and the `Registration form` is required. An example of the new behavior is shown below(Storybook):

Screen Shot 2020-06-05 at 19 20 50

The entire flow is displayed below:

Screen Shot 2020-06-05 at 17


rafaelblink:
Tests are failing and I dunno what's going on.

dougfabris (approved):

PR #430, opened 2020-06-13
fixed small department change errors

vanntile:
Department change doesn't track the right department in local storage so here is a small fix for that.

### DIFF
--- a/src/routes/SwitchDepartment/container.js
+++ b/src/routes/SwitchDepartment/container.js
@@ -28,7 +28,7 @@ export class SwitchDepartmentContainer extends Component {
 
 		if (!room) {
 			const user = await Livechat.grantVisitor({ visitor: { department, token } });
-			await dispatch({ user, alerts: (alerts.push({ id: createToken(), children: I18n.t('Department switched'), success: true }), alerts) });
+			await dispatch({ department, user, alerts: (alerts.push({ id: createToken(), children: I18n.t('Department switched'), success: true }), alerts) });
 			return history.go(-1);
 		}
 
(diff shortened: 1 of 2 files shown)


src/routes/SwitchDepartment/container.js
```diff
@@ -28,7 +28,7 @@ export class SwitchDepartmentContainer extends Component {
 
 		if (!room) {
 			const user = await Livechat.grantVisitor({ visitor: { department, token } });
-			await dispatch({ user, alerts: (alerts.push({ id: createToken(), children: I18n.t('Department switched'), success: true }), alerts) });
+			await dispatch({ department, user, alerts: (alerts.push({ id: createToken(), children: I18n.t('Department switched'), success: true }), alerts) });
```
renatobecker, line 31: This change won't impact anywhere.
As you can see [here](https://github.com/RocketChat/Rocket.Chat.Livechat/blob/develop/src/lib/hooks.js#L68), when you register the department, we update the `iframe` prop, not the root of the object.

src/routes/SwitchDepartment/container.js
```diff
@@ -89,22 +89,22 @@ export const SwitchDepartmentConnector = ({ ref, ...props }) => (
...
-				}}
-				loading={loading}
-				departments={departments.filter((dept) => dept.showOnRegistration && dept._id !== department)}
-				dispatch={dispatch}
-				room={room}
-				alerts={alerts}
-				token={token}
-			/>
-		)}
+				<SwitchDepartmentContainer
```
renatobecker, line 92: Indentation issues.

PR #443, opened 2020-07-09
[NEW] Message character limit feature

oguhpereira:
Depends on https://github.com/RocketChat/Rocket.Chat/pull/18261

The main goal of this PR is to add functionality to add character limit for messages.

ezgif-4-80d563ac6618

#416


renatobecker:
Hey @oguhpereira thanks for contributing, It seems to be a very nice feature!

I have two points to discuss here before we review/merge your code:
- We need the design implemented on the storybook first;
- We need to discuss the design you implemented with our design team, so please share a new gif or a video demo of this feature with higher quality, it's hard to see it on the gif you shared.

I'll be waiting for some news.

Thanks.

oguhpereira:
Hi @renatobecker this is the original [video](https://www.loom.com/share/5bf80229c3454cf5acbf413b2cbde50a).

I went to add the storybook but there seems to be a problem of compatibility of the preact with the version that is in the package.json

renatobecker:
> Hi @renatobecker this is the original [video](https://www.loom.com/share/5bf80229c3454cf5acbf413b2cbde50a).
> 
> I went to add the storybook but there seems to be a problem of compatibility of the preact with the version that is in the package.json

I just watched your video, I noticed that even there is a message length limit, you're allowing the user to type new characters. why?

oguhpereira:
@renatobecker I based it on twitter's UX. I don't prevent the user from entering the message, just block the sending. I'm trying not to create a bad experience when I exceed the character limit. Therefore, even if the user types a message larger than expected, he can adapt to fit the limit.

renatobecker:
> @renatobecker I based it on twitter's UX. I don't prevent the user from entering the message, just block the sending. I'm trying not to create a bad experience when I exceed the character limit. Therefore, even if the user types a message larger than expected, he can adapt to fit the limit.

So, I see two points here:
- if you are just blocking the sending, why do you display the message length limit?
- do you think it's a good UX allowing the users to type hundred words and then tell them are not allowed to send the message because the limit has been reached?

Sry, this doesn't make sense to me.

oguhpereira:
> if you are just blocking the sending, why do you display the message length limit?
do you think it's a good UX allowing the users to type hundred words and then tell them are not allowed to send the message because the limit has been reached?

* For the user to see how many characters passed
* In my opinion, it is a good UX to allow the user to type as much as he wants and then review and delete the excess until it fits the limit

renatobecker:
> > if you are just blocking the sending, why do you display the message length limit?
> > do you think it's a good UX allowing the users to type hundred words and then tell them are not allowed to send the message because the limit has been reached?
> 
> * For the user to see how many characters passed
> * In my opinion, it is a good UX to allow the user to type as much as he wants and then review and delete the excess until it fits the limit

I see.
I understand that for Twitter it may make sense, but here the users aren't blog posting, they're sending messages in a row, which means this isn't the UX our users are expecting for.
If you're willing to implement the feature according to our UI/UX, I will be able to review it.
In addition, there is a conflict in a file, it must be fixed too.

Thanks.

oguhpereira:
> If you're willing to implement the feature according to our UI/UX, I will be able to review it.


@renatobecker, I removed the changes commented above, added a lock to continue typing. Check on [video](https://www.loom.com/share/6c8d724d0712454abebae7badfa9c8de)

renatobecker:
> > If you're willing to implement the feature according to our UI/UX, I will be able to review it.
> 
> @renatobecker, I removed the changes commented above, added a lock to continue typing. Check on [video](https://www.loom.com/share/6c8d724d0712454abebae7badfa9c8de)

Ohh that's awesome!
@oguhpereira what do you think about highlighting the `character counter` when the limit is reached?
Think about it..

oguhpereira:
@renatobecker When text reaches the limit it changes to $color-red. Check on [video](https://www.loom.com/share/49ac85fc74f346f9a698ffe4274a76bd)

src/components/Footer/styles.scss
```diff
@@ -41,6 +41,16 @@
 
 		@include pressable-button(2px);
 	}
+
+	&__remainder, &__remainder-alert  {
```
renatobecker, line 45:
```suggestion
	&__remainder {
		margin-left: 10px;
		min-width: 100px;
		font-weight: bold;

		&--highlight {
			color: $color-red;
			transition: color .2s;
		}
	}
```

src/components/Footer/index.js
```diff
@@ -43,3 +43,13 @@ export const FooterOptions = ({ children }) => (
 		{children}
 	</PopoverMenu>
 );
+
+
+export const ReminderCharacters = ({ className, style = {}, textLenght, limitTextLength }) => (
```
renatobecker, line 48:
```suggestion
export const CharCounter = ({ className, style = {}, textLength, limitTextLength }) => (
	<span
		className={createClassName(styles, 'footer__remainder', { highlight: textLength === limitTextLength }, [className])}
		style={style}
	>
		{textLength} / {limitTextLength}
	</span>
);

```

src/routes/Chat/component.js
```diff
@@ -2,7 +2,7 @@ import { Component } from 'preact';
 
 import { Composer, ComposerAction, ComposerActions } from '../../components/Composer';
 import { FilesDropTarget } from '../../components/FilesDropTarget';
-import { FooterOptions } from '../../components/Footer';
+import { FooterOptions, ReminderCharacters } from '../../components/Footer';
```
renatobecker, line 5:
```suggestion
import { FooterOptions, ChatCounter } from '../../components/Footer';
```

src/routes/Chat/component.js
```diff
@@ -162,14 +172,15 @@ export default class Chat extends Component {
 									<ComposerAction onClick={this.handleUploadClick}>
 										<PlusIcon width={20} />
 									</ComposerAction>
-								)}
+								)}1
```
renatobecker, line 175: ??

renatobecker:
@oguhpereira the storybook is okay now, please pull the last changes on the `develop` branch and try again.
After that, you'll be able to add the new component to the storybook.

oguhpereira:
@renatobecker I removed the number 1, it is not part of the logic, I also added the storybook

PR #461, opened 2020-08-04
[NEW] add support for URI data in audio attachments

murtaza98:
## ISSUE
Currently, the Livechat Widget adds the server Url in front to each of the audio attachment URL. Eg. If the audio-attachment URL is `file-uploads/abc/audio.mp4`, then the widget converts it to `http://localhost:3000/file-uploads/abc/audio.mp4`.

However, this behavior is not required when the data in attachment is in [DATA_URI_SCHEMA](https://en.wikipedia.org/wiki/Data_URI_scheme).

This PR attempts to solve this issue.


src/components/helpers.js
```diff
@@ -116,7 +116,7 @@ export const msgTypesNotRendered = ['livechat_video_call', 'livechat_navigation_
 
 export const canRenderMessage = (message = {}) => !msgTypesNotRendered.includes(message.t);
 
-export const getAttachmentUrl = (url) => `${ Livechat.client.host }${ url }`;
+export const getAttachmentUrl = (url) => url && url.startsWith('data:') ? url : `${ Livechat.client.host }${ url }`;
```
renatobecker, line 119:
```suggestion
export const getAttachmentUrl = (url) => (url && url.startsWith('data:') ? url : `${ Livechat.client.host }${ url }`);
```

Issue #470, opened 2020-08-13
HTML is not escaped on paste/drop text


Shailesh351:
I copied html of message list from dev tools and paste it in composer. It looks as below.

![localhost_8080_lc](https://user-images.githubusercontent.com/18264684/90106872-a6514200-dd65-11ea-9735-a335afb2cdf5.png)

PR #471, opened 2020-08-13
Fix Escaping HTML on paste/drop Text

Shailesh351:
Closes https://github.com/RocketChat/Rocket.Chat.Livechat/issues/470


MartinSchoeler (approved):
Good catch! Looks good to me!

Issue #232, opened 2019-05-13
Only first trigger working


Simonell:
When we have more than one trigger, then only first is working.
Rest not working on another URL's. We are trying to add LiveChat on multiple websites with "Visitor Page URL" condition.

renatobecker:
@Simonell can you share your configuration, please?
Thanks.

Simonell:
![Screenshot_2019-05-13 (2) eVolpe](https://user-images.githubusercontent.com/2706043/57630121-4cc7f980-759d-11e9-90f6-b35e09c58a47.png)
I have 3 triggers like this.
My Rocket Chat version is 1.0.2.

renatobecker:
Could you please list your website URLs?

Simonell:
Regexes looks like on screenshot, but with different suffixes (.de/.com/.pl)

wt., 14 maj 2019 o 16:11 Renato Becker <notifications@github.com>
napisał(a):

franckadil:
I tested and can reproduce. Only the first trigger works, When adding for each page a trigger (website URL) with this regex: .pagename.

thigas88:
Triggers not working for default visitor message

![image](https://user-images.githubusercontent.com/704549/66076840-0c115280-e535-11e9-9d71-cca227d10982.png)

config

![image](https://user-images.githubusercontent.com/704549/66076920-319e5c00-e535-11e9-8e05-a45c59e29636.png)

etlam:
Still not working v.3.9.0.  
Only the first activated trigger is working.

JoelBermudezG:
+1
Not yet implemented, and key for custom messages according to the context of the page.

PR #537, opened 2020-11-29
[FIX] Only first trigger working #232

etlam:
Only the first trigger of type page-url was working. Requests got deleted before checking the second trigger.

### DIFF
--- a/src/lib/triggers.js
+++ b/src/lib/triggers.js
@@ -160,9 +160,9 @@ class Triggers {
 							const hrefRegExp = new RegExp(condition.value, 'g');
 							if (request.location.href.match(hrefRegExp)) {
 								self.fire(trigger);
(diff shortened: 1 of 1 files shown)


rafaelblink:
@etlam Hi dude, thanks for your PR, I'm trying to test your implementation but still receiving only the first trigger message, could you test again?

etlam:
@rafaelblink How did you test the behaviour?  
I did not found any way to test the change, I just did it by reading the source code and checking possibilities why other triggers do not work

cauefcr:
Unfortunately the bug is a bit deeper than just changing this variable, I've started a branch to work on this issue, here it is: https://github.com/RocketChat/Rocket.Chat.Livechat/tree/fix/broken-triggers

Issue #403, opened 2020-05-05
Rocket.Chat 3.0.12 Livechat issue about Korean IME


chojunki:
### Description:

When you type in some messages in Korean through livechat windows, the Content get corrupted.

### Steps to reproduce:

1. Open livechat on a Korean IME enabled PC with New Edge or Chrome browser. Then It will look like this:
![1](https://user-images.githubusercontent.com/24218869/78880151-ee96e580-7a8f-11ea-8dfd-7184d1cffed3.png)

2. Type any messages in Korean in the field showing 'Type your message here'
3. At first keystroke, the 'Type your message here' will disappear but IME will stop composing keystrokes to assemble a syllable like thks:
![2](https://user-images.githubusercontent.com/24218869/78880334-36b60800-7a90-11ea-8139-06aff5a3c42e.png)
4. After this, The text will be entered normally.

### Expected behavior:
At first keystroke, the 'Type your message here' should disappear and keystroke composion must go normally.

### Actual behavior:
The only first keystroke that makes 'Type your message here' only removes the message and keystroke composition feature just breaks up.

I think this is because of the IME characteristics related with onKeyDown, onKeyUp and onKeyPress event. I found that I should not use onKeyUp event with Korean IME because that event will break up character composition feature of Korean IME.

### Server Setup Information:
This is just a client issue.

- Version of Rocket.Chat Server: 3.0.12
- Operating System: CentOS 7.4
- Deployment Method: manual installation
- Number of Running Instances: 1
- DB Replicaset Oplog: enabled
- NodeJS Version: 12.14.0
- MongoDB Version: 4.2.0, but this is not the case this time.

### Client Setup Information

- Desktop App or Browser Version: New Edge and Chrome. Both current version.
- Operating System: WIndows 10 Pro Korean, 19H2 with all updates

luixxiul:
@renatobecker FWIW the issue can be avoided by replacing `<div>` whose class is `composer__input__ubV5B` with a `<input>`.

![livechat3](https://user-images.githubusercontent.com/3362943/81268073-fae27400-9036-11ea-9b1d-af1dcfab263e.gif)

The paper plane icon to send a message stop appearing so another modification would be required.

luixxiul:
The issue is reproduced on Chinese IME too. The issue is common among CJK environment.

![livechat5](https://user-images.githubusercontent.com/3362943/81271486-ad1c3a80-903b-11ea-9cb1-341dc0362872.gif)

[...]

PR #538, opened 2020-12-04
[FIX] Rocket.Chat 3.0.12 Livechat issue about Korean IME #403

ian902792:
This way can fix IME input method problem(Such as, Korean、Japanese、Chinese).
Because preact.js has SyntheticEvent issue.
Therefore, oncompositionstart & oncompositionend are all lowercase.
[Reference](https://github.com/preactjs/preact/issues/1978)

### DIFF
--- a/src/components/Composer/index.js
+++ b/src/components/Composer/index.js
@@ -56,13 +56,17 @@ const replaceCaret = (el) => {
 	}
 };
 
+let inputLock = false;
+
 export class Composer extends Component {
 	handleRef = (el) => {
 		this.el = el;
 	}
 
 	handleInput = (onChange) => () => {
-		onChange && onChange(this.el.innerText);
+		if (!inputLock) {
+			onChange && onChange(this.el.innerText);
+		}
 	}
 
 	handleKeypress = (onSubmit) => (event) => {
@@ -246,6 +250,16 @@ export class Composer extends Component {
 						onClick: this.handleClick,
 					}
 				)}
+
+				oncompositionstart={(e)=>{
+					inputLock = true;
+				}}
+
+				oncompositionend={(e)=>{
+					inputLock = false;
+					onChange && onChange(this.el.innerText);
+				}}
+
 				className={createClassName(styles, 'composer__input')}
 			/>
 			{post}


ian902792:
Contributor License Agreement page show "You have agreed to the CLA for RocketChat/Rocket.Chat.Livechat"
but GitHub comment's CLA status show " not signed yet".
So...?

sctEdwin:
May I asked when will this be merged?
Japanese IME conversion is also fixed by this pull request.

rafaelblink (approved):

src/components/Composer/index.js
```diff
@@ -56,13 +56,17 @@ const replaceCaret = (el) => {
 	}
 };
 
+let inputLock = false;
```
renatobecker, line 59: Why did you implement a global variable rather than an internal state variable?

src/components/Composer/index.js
```diff
@@ -56,13 +56,17 @@ const replaceCaret = (el) => {
...
+
 export class Composer extends Component {
 	handleRef = (el) => {
 		this.el = el;
 	}
 
 	handleInput = (onChange) => () => {
-		onChange && onChange(this.el.innerText);
+		if (!inputLock) {
```
renatobecker, line 67:
```suggestion
		if (inputLock) {
			return;
		}
		
		onChange && onChange(this.el.innerText);
```

src/components/Composer/index.js
```diff
@@ -246,6 +250,16 @@ export class Composer extends Component {
 						onClick: this.handleClick,
 					}
 				)}
+
+				oncompositionstart={(e)=>{
```
renatobecker, line 254:
```suggestion
				onCompositionStart={(e)=>{
```
Use camelCase to write event names: https://reactjs.org/docs/events.html#composition-events

renatobecker (changes requested):

tiagoevanp:
Thank you @ian902792 for the suggestion!

In the future we will improve the composer as a more complex input component. For now is more concise to deal with the problem in that way:  https://github.com/RocketChat/Rocket.Chat.Livechat/pull/674

PR #553, opened 2021-02-11
[FEATURE] New trigger messages style

### LINKED ISSUE #244: On mobile Livechat covers the whole screen, making it intrusive for triggers.

rafaelblink:
CLOSES #244 

When browsing websites on mobile, visitors are facing a very bad user experience when livechat trigger messages are fired.

### Before
image

#### Single
image

#### Multiple
image

### Desktop
image

### Mobile
image


dougfabris (changes requested):
@rafaelblink really nice improvements! I found some points to discuss: 

1. I think we shouldn't render empty triggers:
image
2. To improve more the second point, maybe we can add a UI validation not allowing the creation of triggers with the message field empty
3. It's not possible to run the storybook. I think it's something wrong in  @rocket.chat/ui-kit package version. It fires an error on `src/components/uiKit/message/index.js`

renatobecker (changes requested):
Screen Shot 2021-02-18 at 5 56 36 PM

The new component is not following the design system(figma).

Screen Shot 2021-02-18 at 5 58 14 PM

rafaelblink:
> @rafaelblink really nice improvements! I found some points to discuss:
> 
> 1. I think we shouldn't render empty triggers:
>    image
> 2. To improve more the second point, maybe we can add a UI validation not allowing the creation of triggers with the message field empty
> 3. It's not possible to run the storybook. I think it's something wrong in  @rocket.chat/ui-kit package version. It fires an error on `src/components/uiKit/message/index.js`

PR created to solve this: https://github.com/RocketChat/Rocket.Chat/pull/20827

About storybook, its running here, I'm running with `yarn storybook`

image

dougfabris (approved):

PR #554, opened 2021-02-16
[IMPROVE] System messages style

rafaelblink:
The current version of the livechat widget is not dealing well with system messages, they're not user-friendly.

## Before
image

## After
image


renatobecker:
The agent avatar is not working properly after a system message is sent in the chat.
The image attached in the description shows that, so we need to fix it.

rafaelblink:
> The agent avatar is not working properly after a system message is sent in the chat.
> The image attached in the description shows that, so we need to fix it.

I'm trying to see whats going on, @dougfabris, could you help to find this?

dougfabris (approved):

PR #559, opened 2021-03-08
[FIX] Invalid font size for hiragana and katakana

sctEdwin:
Two problem experienced in implementing Japanese livechat.
1. hiragana and katakana only text input by user is reflected in large size.
Before fix:
image

After fix:
image

2. ime conversion problem cause some Japanese to experience early conversion before finish inputting
duplicate with  https://github.com/RocketChat/Rocket.Chat.Livechat/pull/538


rafaelblink:
Hi @sctEdwin how are you man? Thanks for you contribution.

I'm reviewing your PR, and I've noticed that you removed the size of the `transformEmojisToLargeSize`, that `font-size` is required to transform only `emojis` messages into a bigger size:

image

This is why that `font-size: 2rem` must be there, we cannot remove.

The main thing here, is because the method `isOnlyEmoji` method is recognizing japanese characters as emojis and we need to change that.

src/components/Composer/index.js
```diff
@@ -56,13 +56,19 @@ const replaceCaret = (el) => {
...
+
+
 export class Composer extends Component {
 	handleRef = (el) => {
 		this.el = el;
 	}
 
 	handleInput = (onChange) => () => {
+		if (!inputlock){
 		onChange && onChange(this.el.innerText);
```
rafaelblink, line 70:
```suggestion
			onChange && onChange(this.el.innerText);
```

src/components/Composer/index.js
```diff
@@ -56,13 +56,19 @@ const replaceCaret = (el) => {
...
+let inputLock = false;
+
+
 export class Composer extends Component {
 	handleRef = (el) => {
 		this.el = el;
 	}
 
 	handleInput = (onChange) => () => {
+		if (!inputlock){
```
rafaelblink, line 69:
```suggestion
		if (!inputLock) {
```

rafaelblink (changes requested):

sctEdwin:
> 
> 
> Hi @sctEdwin how are you man? Thanks for you contribution.
> 
> I'm reviewing your PR, and I've noticed that you removed the size of the `transformEmojisToLargeSize`, that `font-size` is required to transform only `emojis` messages into a bigger size:
> 
> image
> 
> Your solution solve the problem but `only emojis messages` are not being rendered as big size.
> 
> image
> 
> This is why that `font-size: 2rem` must be there, we cannot remove.
> 
> The main thing here, is because the method `isOnlyEmoji` method is recognizing japanese characters as emojis and we need to change that.

I exclude hiragana and katakana from being treated as emoji so no change is necessary for `transformEmojisToLargeSize` now. I checked the behavior for IE, Firefox, Chrome and Edge and all seems working well, hopefully you can accept this commit as hiragana and katakana are not emoji.
seemsOK

Reference for hiragana and katakana unicode codepoint:
[hiragana](http://www.unicode.org/charts/PDF/U3040.pdf)
[katakana](https://www.unicode.org/charts/PDF/U30A0.pdf)

src/components/Composer/index.js
```diff
@@ -246,6 +252,16 @@ export class Composer extends Component {
 						onClick: this.handleClick,
 					}
 				)}
+
+				oncompositionstart={(e)=>{
```
rafaelblink, line 256:
```suggestion
				onCompositionStart={(e)=>{
```

src/components/Composer/index.js
```diff
@@ -246,6 +252,16 @@ export class Composer extends Component {
...
+					inputLock = true;
+				}}
+
+				oncompositionend={(e)=>{
```
rafaelblink, line 260:
```suggestion
				onCompositionEnd={(e)=>{
```

rafaelblink (changes requested):
Hi @sctEdwin

Please check followed documentation and see that events names are `camelCase`.

https://reactjs.org/docs/events.html#composition-events

src/components/Composer/index.js
```diff
@@ -56,13 +56,19 @@ const replaceCaret = (el) => {
 	}
 };
 
+
+let inputLock = false;
```
rafaelblink, line 60: We could transform this into a state instead a global variable.

rafaelblink (changes requested):

renatobecker:
@sctEdwin can you please fix the conflicting file? 
Thanks.

src/components/Composer/index.js
renatobecker: @sctEdwin You marked this comment as resolved but you didn't change it. Why?

renatobecker (changes requested):

src/components/Composer/index.js
sctEdwin: @rentaobecker
Sorry I am new to this, did my changes address the issue?

sctEdwin:
@renatobecker sorry it is a mess so I rebase it with the current develop branch in main repository and add the changes again.

src/components/Composer/index.js
```diff
@@ -63,6 +63,9 @@ export class Composer extends Component {
 	}
 
 	handleInput = (onChange) => () => {
+		if (inputLock) {
```
renatobecker, line 66: This is not a state, the variable is not declared as a component state property.

renatobecker (changes requested):

src/components/Composer/index.js
sctEdwin: @renatobecker 
Does my latest commit fulfill the requirement?
If not, can you provide a bit hint on that?
The build and behavior seems alright.

sctEdwin: @renatobecker
I dont know how to implement a state variable as I have to use setState() in onCompositionStart() and onCompositionEnd(),
which is part of the render function. I read about not using setState() in render function but I dont see a way to avoid it.
[do not use setState() in render](https://stackoverflow.com/a/51677449),
I currently call the function handleInputLock to setState but I doubt it is keeping render function pure.
Any suggestion or hint?

src/components/Composer/index.js
```diff
@@ -247,6 +260,18 @@ export class Composer extends Component {
 						onClick: this.handleClick,
 					}
 				)}
+		
+
+				onCompositionStart={(e)=>{
```
rafaelblink, line 265:
```suggestion
				onCompositionStart={() => {
```

src/components/Composer/index.js
```diff
@@ -247,6 +260,18 @@ export class Composer extends Component {
...
+					this.handleInputLock(true);
+				}}
+
+				onCompositionEnd={(e)=>{
```
rafaelblink, line 269:
```suggestion
				onCompositionEnd={() => {
```

src/components/Composer/index.js
```diff
@@ -228,7 +235,13 @@ export class Composer extends Component {
 		return 0;
 	}
 
+	handleInputLock(isNot){
```
rafaelblink, line 238:
```suggestion
	handleInputLock(isNot) {
```

src/components/Composer/index.js
```diff
@@ -147,6 +151,9 @@ export class Composer extends Component {
 
 	constructor(props) {
 		super(props);
+		this.state = {  
+				inputLock: false
```
rafaelblink, line 155:
```suggestion
		inputLock: false
```

src/components/Composer/index.js
```diff
@@ -58,11 +58,15 @@ const replaceCaret = (el) => {
 };
 
 export class Composer extends Component {
+
```
rafaelblink, line 61:
```suggestion
```

rafaelblink (changes requested):
Hi @sctEdwin, please look at some changes requests that I made, ins't complex changes, please take attention on your code indentation, if possible, install `ESLint` on your code editor.

rafaelblink:
Hi @sctEdwin, the PR isn't passing on the lint, please, see whats going on here.
https://github.com/RocketChat/Rocket.Chat.Livechat/pull/559/checks?check_run_id=2348204157

sctEdwin:
> Hi @sctEdwin, the PR isn't passing on the lint, please, see whats going on here.
> https://github.com/RocketChat/Rocket.Chat.Livechat/pull/559/checks?check_run_id=2348204157

@rafaelblink I installed ESLint onto Vim and fixed the error shown, it seems that the PR passed the test.

rafaelblink (approved):

src/components/Composer/index.js
```diff
@@ -228,7 +234,13 @@ export class Composer extends Component {
 		return 0;
 	}
 
+	handleInputLock(isNot) {
```
renatobecker, line 237: What does `isNot` mean? Why not something like `locked` ?

rafaelblink (approved):

dougfabris (approved):

Issue #561, opened 2021-03-16
Incomplete Email Validation


Deepak-learner:
Description:
In Rocket.Chat.Livechat app , Email validation is wrong in offline registration form . If email  contain  only "@"  without " . " (dot)  in email field text box with valid email address format  , "send" button in form is not activated i.e invalid email address but the email addresses of format example@localhost is a valid type  . For valid email address format , visit this site : https://en.wikipedia.org/wiki/Email_address#Examples

Steps to reproduce:
Go to "Offline Registration Form" in Livechat widget  .
Enter email without "."(dot) in email text box.

Expected behavior:
"send" button should be activated .

Actual behavior:
"send" button is not activated.

Server Setup Information:
Operating System: Ubuntu

Client Setup Information
Desktop App or Browser Version: Chrome version 87
Operating System: Ubuntu

Additional Context 
Actual 
<img width="1263" alt="Screenshot 2021-03-16 at 2 16 04 PM" src="https://user-images.githubusercontent.com/56799414/111281498-00e7af00-8663-11eb-8506-ad00aca0f5d5.png">

"send" button is not activated .

PR #563, opened 2021-03-16
[IMPROVE] Allow more email domains on Livechat registration

Deepak-learner:
#561 [FIX] Email-validation
Done email validation in offline registration form .
Screenshot 2021-03-16 at 2 19 41 PM
'send' button is activated for correct email address .


Deepak-learner:
@rafaelblink please review .

Deepak-learner:
@rafaelblink @murtaza98 please review.

murtaza98:
@Deepak-learner I don't think an email is now considered valid if it doesn't have a TLD.

You can also also check this on any online validation website. I used this eg: :point_down: 

image

Deepak-learner:
In rocket chat web app , it's accepting "test@test" email address and saving the details in contact list .
Screenshot 2021-03-18 at 2 31 02 PM
Also in this link : https://en.wikipedia.org/wiki/Email_address#Examples , check 'Valid email addresses' section  , its mention 
"admin@mailserver1" is valid email address . 
@murtaza98 , @rafaelblink can you please confirm once , particular this email address format is valid or not  as it is  also use in rocket chat web desktop app as shown in above image.

cauefcr:
Hey people, this PR and its discussion got me in the middle of a RFC rabbit hole, and from what i can tell, [here](https://datatracker.ietf.org/doc/html/rfc2822#section-3.4.1) is the first grammar definition of an e-mail address, as addr-spec, but a few following RFCs made changes to it, like [here](https://datatracker.ietf.org/doc/html/rfc5322#section-3.4.1), allowing changes in implementation, to guarantee it's working on the specific app's context. Internationalization is another concern, as I've found no regex capable of dealing with all internationalization problems. According to [this](https://www.regular-expressions.info/email.html) website is not even possible to trust the standard regex given by IETF [here](https://tools.ietf.org/id/draft-seantek-mail-regexen-03.html#rfc.section.5). As the official recommended regex engine won't even work in Javascript, the best suggestion I have for now is the regex below, found [here](https://emailregex.com/), and that's not even talking about security implications of over-matching or frustration from users in under-matching.
```regex
/^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
```
I don't believe there is a right answer here, what do you all think?

KevLehman:
Well, we should use the one provided by IETF (as its "standard"). Anyways, we should be consistent 🤔 so let's give this a closer look

cauefcr:
Hey, after raising the issue of regex e-mail validation with our team the great @KevLehman made changes on the main product, which i have copied over to this repo, on PR https://github.com/RocketChat/Rocket.Chat.Livechat/pull/693, thank you for reporting the problem!

bastelix:
image

I get this every try .. cannot disable email and no email will accepted. Very buggy.

bubble0h7:
> I get this every try .. cannot disable email and no email will accepted. Very buggy.

@bastelix I managed to fix this issue by setting the "Email Address to Send Offline Messages" field in the `Omnichannel > Livechat Appearance > Omnichannel offline` settings. Not the most helpful error message though.

Issue #351, opened 2020-01-08
LiveChat Widget doesn't always load messages


wreiske:
1. Start a new livechat
2. Send a message
3. Reload the screen a few times and see if the message is there

Seems like every once in awhile the messages won't show, or they only show for a second and then they go away.

@renatobecker I've sent you a GIF privately of the error occurring.

grestrepo:
someone resolved this problem?

PR #571, opened 2021-03-23
[FIX] many-request-error

Deepak-learner:
[FIX] issue #351 
when I reproduce the issue and check in inspect , network section , I get below result:
Screenshot 2021-03-24 at 3 06 25 AM

So , I fix this issue by showing this error to user , and ask him to wait as page will reload after given time in error message . For more clarity check this out:

https://user-images.githubusercontent.com/56799414/112222638-d42d2c00-8c4e-11eb-82e1-e47d63df6ec7.mov

### DIFF
--- a/src/lib/room.js
+++ b/src/lib/room.js
@@ -162,15 +162,38 @@ export const loadMessages = async () => {
 	}
 
 	await store.setState({ loading: true });
-	const rawMessages = (await Livechat.loadMessages(rid)).concat(previousMessages);
-	const messages = (await normalizeMessages(rawMessages)).map(transformAgentInformationOnMessage);
+	try {
+		const rawMessages = (await Livechat.loadMessages(rid)).concat(previousMessages);
+		const messages = (await normalizeMessages(rawMessages)).map(transformAgentInformationOnMessage);
+		await initRoom();
+		await store.setState({ messages: (messages || []).reverse(), noMoreMessages: false, loading: false });
 
-	await initRoom();
-	await store.setState({ messages: (messages || []).reverse(), noMoreMessages: false, loading: false });
-
-	if (messages && messages.length) {
-		const lastMessage = messages[messages.length - 1];
-		await store.setState({ lastReadMessageId: lastMessage && lastMessage._id });
+		if (messages && messages.length) {
+			const lastMessage = messages[messages.length - 1];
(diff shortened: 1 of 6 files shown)


Deepak-learner:
@renatobecker please review.

src/lib/room.js
```diff
@@ -162,15 +162,38 @@ export const loadMessages = async () => {
...
+		await initRoom();
+		await store.setState({ messages: (messages || []).reverse(), noMoreMessages: false, loading: false });
+
+		if (messages && messages.length) {
+			const lastMessage = messages[messages.length - 1];
+			await store.setState({ lastReadMessageId: lastMessage && lastMessage._id });
+		}
+	}
+	catch(e) {
+		const waitTime = parseInt(e.data.error.match(/^.*You must wait ([0-9]*) seconds.*$/)[1]);
```
rafaelblink, line 177: Please set this message on i18n files.

src/lib/room.js
```diff
@@ -162,15 +162,38 @@ export const loadMessages = async () => {
...
+		const decrementCounter = async () => {
+			if(window.loadMessagesTimeout) {
+				clearInterval(window.loadMessagesTimeout);
+			}
+			window.loadMessagesTimeout = setInterval(() => {
+				let tmpWaitTime = store.state.loadWaitTime;
+				if(tmpWaitTime) {
+					store.setState({loadWaitTime: tmpWaitTime - 1, loadingMessage: `Error, too many requests. Please slow down. You must wait for ${tmpWaitTime} seconds.` });
```
rafaelblink, line 185: Please set this message on i18n files.

src/lib/room.js
rafaelblink, line 177: Please set this message on i18n files.

src/lib/room.js
Deepak-learner: according to your comment, does I need to set that message in all i18n files. Like I have to translate that message to all other languages and set it to respective i18n json files. Right? And if this is the case how I can identify which i18n json file belong to which language. Is there any fast way to do this or I have to manually check each i18n json file to check which language it belongs to. Looking forward to your response. Thanks.

rafaelblink: please check the `i18n` command on `package.json` files.

You'll use with `I18n.t('YOUR TEXT')` and run the command, that command will create the line on i18n files.

PR #610, opened 2021-05-21
[IMPROVE] Open links in another tab on Livechat widget

billynoah:
prevent target attribute from being removed by sanitizer


MartinSchoeler (commented):
Looks good to me

MartinSchoeler (approved):

PR #617, opened 2021-06-27
add callnotification,callStartmsg,iframe code

Deepak-learner:
Whenever a call is made to the visitor side by agent, the call notification dialog box will appear at the visitor side with an accept and reject button. If a visitor clicks on the accept button, the video call will start in the iframe. And if the user clicks on the decline button the dialog box will be closed. 

Below I attach a video link showing the implementation

Web view : 
https://drive.google.com/file/d/14Gn4noVgeqM742kWaI6j5Obk2oWsCSAG/view?usp=sharing

Mobile View:
https://drive.google.com/file/d/1ukv0CxFPEYwDG1LnyyrijrUVe3P6kYWy/view?usp=sharing


murtaza98:
@Deepak-learner Couple of points here
1. The following icons are missing from the `Accept` and `Decline` buttons. Please add them as well
image
2. The position of various components within your implementation doesn't seem to follow the original design. Take a look at the following comparison between your implementation and the original design and notice the coloured rectangles I drew on both of them. Relatively both of them should be the same even but in your implementation it seems different. There is definitely some room for improvement over here.  
image
3. Could you please record a demo for the mobile devices as well and share it with us? You can use the chrome developer for this. Also for the desktop demo, could you modify your local setup to make sure the livechat doesn't cover the entire screen something like this. You'd just have to embed the livechat code within some demo html file in order to do this.
image

src/routes/Chat/component.js
```diff
@@ -165,6 +166,9 @@ export default class Chat extends Component {
 							autoFocus={true}
 						/>}
 					</div>
+					<div>
+						{(messages[messages.length - 1] ? messages[messages.length - 1].t === 'jitsi_call_started' : messages[messages.length - 1]) ? <CallNotification rid={(messages[messages.length - 1])} /> : null}
```
murtaza98, line 170: This is temporary right? I was under the impression that we'd be relying on the video call event instead of the message type to identify if the call has started

src/components/Messages/MessageList/livechatCall.js
```diff
@@ -0,0 +1,60 @@
...
+import { Button } from '../../Button';
+import { Screen } from '../../Screen';
+import { createClassName, getAvatarUrl } from '../../helpers';
+import { parseDate } from '../MessageTime/index';
+import styles from './styles.scss';
+
+// get roomid(GetRoomId(rid)) and put rid id in iframe link
+// <iframe style="height:80%" src="localhost:3000/livechat/roomId"></iframe>
+// For example, for now I have add link of jitsi and its working fine, similarly we can add link of our video/audio call implemented with webrtc.
+const GetRoomId = (rid) => (
```
murtaza98, line 14: :man_shrugging: The function name is `GetRoomId` but its functionality is to display the iframe :confused: ?

src/components/Messages/Message/index.js
```diff
@@ -141,5 +142,6 @@ export const Message = memo(({
 			})}
 		</MessageContent>
 		{!compact && !message.t && <MessageTime normal={!me} inverse={me} ts={ts} />}
+		{message.t === 'jitsi_call_started' && message.callStatus === 'accept' ? <ShowCallTime stime={ts} /> : null}
```
murtaza98, line 145: @Deepak-learner Can we reuse the existing system messages instead of creating a new Component for the same functionality? You can refer to [this PR](https://github.com/RocketChat/Rocket.Chat.Livechat/pull/554) for more info

src/components/Messages/MessageList/livechatCall.js
```diff
@@ -0,0 +1,62 @@
...
+				<div className={createClassName(styles, 'call')}>
+					<div className={createClassName(styles, 'avatar')}>
+						<Avatar
+							src={getAvatarUrl(props.rid.u.username)}
+						/></div>
+					Incoming video Call
+					<div className={createClassName(styles, 'btn')}>
+						<Button onClick={declineClick} className={createClassName(styles, 'btn1')}> <PhoneDecline width={20} height={20} /> Decline</Button>
+						<Button onClick={acceptClick} className={createClassName(styles, 'btn2')} > <PhoneAccept width={20} height={20} /> Accept</Button></div></div></Screen.Content>) : null}
+			{isframe ? (<DisplayIframe session={props.rid} />) : null }
```
murtaza98, line 60:
```suggestion
			{isframe ? (<DisplayCallIframe session={props.rid} />) : null }
```

src/components/Messages/MessageList/livechatCall.js
```diff
@@ -0,0 +1,62 @@
...
+import { Button } from '../../Button';
+import { Screen } from '../../Screen';
+import { createClassName, getAvatarUrl } from '../../helpers';
+import { parseDate } from '../MessageTime/index';
+import styles from './styles.scss';
+
+// get roomid(DisolayIframe(rid)) and put rid id in iframe link
+// <iframe style="height:80%" src="localhost:3000/livechat/roomId"></iframe>
+// For example, for now I have add link of jitsi and its working fine, similarly we can add link of our video/audio call implemented with webrtc.
+const DisplayIframe = (rid) => (
```
murtaza98, line 16:
```suggestion
const DisplayCallIframe = (rid) => (
```

src/components/Messages/MessageList/livechatCall.js
```diff
@@ -0,0 +1,62 @@
...
+import { Avatar } from '../../Avatar';
+import { Button } from '../../Button';
+import { Screen } from '../../Screen';
+import { createClassName, getAvatarUrl } from '../../helpers';
+import { parseDate } from '../MessageTime/index';
+import styles from './styles.scss';
+
+// get roomid(DisolayIframe(rid)) and put rid id in iframe link
+// <iframe style="height:80%" src="localhost:3000/livechat/roomId"></iframe>
+// For example, for now I have add link of jitsi and its working fine, similarly we can add link of our video/audio call implemented with webrtc.
```
murtaza98, line 15: Please make sure to remove these comments in your final PR. We can't keep this PR in production

src/components/Messages/MessageList/livechatCall.js
```diff
@@ -0,0 +1,62 @@
...
+		<div>
+			{ show ? (<Screen.Content nopadding>
+				<div className={createClassName(styles, 'call')}>
+					<div className={createClassName(styles, 'avatar')}>
+						<Avatar
+							src={getAvatarUrl(props.rid.u.username)}
+						/></div>
+					Incoming video Call
+					<div className={createClassName(styles, 'btn')}>
+						<Button onClick={declineClick} className={createClassName(styles, 'btn1')}> <PhoneDecline width={20} height={20} /> Decline</Button>
```
murtaza98, line 58:
```suggestion
						<Button onClick={declineClick} className={createClassName(styles, 'declineButton')}> <PhoneDecline width={20} height={20} /> Decline</Button>
```
Try to name the classes properly based on the class function. Please do the same for others

src/components/Messages/MessageList/styles.scss
```diff
@@ -33,3 +33,86 @@
...
+
+	text-align: center;
+
+	color: #ffffff;
+
+	border: 1px solid black;
+
+	background: #1f2329;
+
+	font-family: Inter;
```
murtaza98, line 59: Lets not override the global font family. We should try to keep the font consistent across entire app

src/components/Messages/MessageList/livechatCall.js
```diff
@@ -0,0 +1,62 @@
...
+						<Button onClick={acceptClick} className={createClassName(styles, 'btn2')} > <PhoneAccept width={20} height={20} /> Accept</Button></div></div></Screen.Content>) : null}
```
murtaza98, line 59: There's no padding between icon and the `Accept` and `Decline` text. Could you please add that?

src/components/Messages/MessageList/livechatCall.js
```diff
@@ -0,0 +1,62 @@
...
+
+	return (
+		<div>
+			{ show ? (<Screen.Content nopadding>
+				<div className={createClassName(styles, 'call')}>
+					<div className={createClassName(styles, 'avatar')}>
+						<Avatar
+							src={getAvatarUrl(props.rid.u.username)}
+						/></div>
+					Incoming video Call
```
murtaza98, line 56: You're hardcoding this text here. This is not a good idea. You need to wrap it around `I18n.t()` to provide translations for it

murtaza98: Please do the same for the Accept and Decline button

src/components/Messages/MessageList/styles.scss
```diff
@@ -33,3 +33,86 @@
...
+
+	margin-bottom: 1px;
+	padding: 10px;
+	padding-top: 0;
+
+	color: white;
+	justify-content: center;
+}
+
+.avatar {
```
murtaza98, line 79: On the original design I think the avatar looks a bit bigger? Could you please check?

murtaza98 (changes requested):
Hello @Deepak-learner Overall this looks good. I left a couple of comments, please take a look at them. Apart from that, upon testing I feel like the jitsi UI is not working properly within our small iframe layout. Eg: on small devices, I'm not able to see all the buttons. also I wasn't able to enter Display name properly on jitsi while starting the call bcs my mobile keyboard was overlapping with the text box. 
So i was thinking maybe for jitsi when the user clicks on the Accept button, for both Desktop and mobile, we should redirect them to a new tab instead with the jitsi link. WDYT? I'd like to know your thoughts on this one.

One more thing, we're currently not handling the browser permission properly over here. Ideally upon Clicking the `Accept` button, we should check if the user has approved the necessary permissions or not and then handle all these scenerios
- If they haven't accepted the permission we should show the popup asking for them to accept the permission. This could be the default browser popup which we get on chrome.
- If the user has declined the permission, we shouldn't start the call, instead show an error message on the livechat something like this (PS: in the example it is a warning message, but we display error messages similarly)
image
- If the user accepts the permission (or has already approved the permissions) then we should open the iframe or redirect the user to the jitsi or webrtc call.

Please let me know if you have any doubts or perhap some new suggestions on these items. Thanks and great work so far. Congrats :smile:

Deepak-learner:
Hii @murtaza98 . Thanks for the feedback.I will work on your above comments and will get back to you. For jitsi, I think we can go with your suggested approach. Regarding browser permission like to access camera and microphone,  I have this point in mind and I think this functionality needs to be added at rc-core repository side.

src/components/Messages/MessageList/livechatCall.js
```diff
@@ -0,0 +1,61 @@
...
+import { Screen } from '../../Screen';
+import { createClassName, getAvatarUrl } from '../../helpers';
+import { parseDate } from '../MessageTime/index';
+import styles from './styles.scss';
+
+
+const DisplayCallIframe = (rid) => (
+	<Screen.Content nopadding>
+		<div className={createClassName(styles, 'iframe')}>
+			<iframe style='height:100%' src='https://meet.jit.si/RocketChatXLKhe6QE6dyRLtTkX'{...rid} />
```
murtaza98, line 18: Why is the following value hard-coded here - https://meet.jit.si/RocketChatXLKhe6QE6dyRLtTkX ? The livechat should get the jitsi server info from the core.

src/components/Messages/MessageList/livechatCall.js
```diff
@@ -0,0 +1,66 @@
...
+
+
+export const CallNotification = (props) => {
+	const [isframe, setIframe] = useState(false);
+	const [show, setShow] = useState(true);
+
+	const acceptClick = () => {
+		setShow(!{ show });
+		if (props.rid.t === 'jitsi_call_started') {
+			window.open(`https://meet.jit.si/xlkhe6qe6dyrlttkx${ props.rid.rid }`);
```
murtaza98, line 42: Jitsi server link is harcoded here as well. Please change this

src/routes/Chat/component.js
```diff
@@ -165,6 +166,9 @@ export default class Chat extends Component {
 							autoFocus={true}
 						/>}
 					</div>
+					<div>
+						{(messages[messages.length - 1] ? messages[messages.length - 1].t === 'webRTC_call_started' || messages[messages.length - 1].t === 'jitsi_call_started' : messages[messages.length - 1]) ? <CallNotification rid={(messages[messages.length - 1])} /> : null}
```
murtaza98, line 170: TBH, I didn't like what we're doing here. The logic for determining if we should show the call notification shouldn't be present in this component bcs its not the responsibility of this component. We should actually pass a boolean prop to this component which would indicate if the callNotification component should be visible or not. Also I think we should be storing this boolean prop value within the Store itself (src/store/index.js)

src/components/Messages/Message/index.js
```diff
@@ -141,5 +142,6 @@ export const Message = memo(({
 			})}
 		</MessageContent>
 		{!compact && !message.t && <MessageTime normal={!me} inverse={me} ts={ts} />}
+		{message.t === 'webRTC_call_started' && message.callStatus === 'accept' ? <ShowCallTime stime={ts} /> : null}
```
murtaza98, line 145: This should be changed to system messages defined [here](https://github.com/RocketChat/Rocket.Chat.Livechat/pull/617/files#diff-bb3347eeb1e2807319f728d4892a5edba39d39eab105e29fd6c120d2820023e1R84). Also it seems here we're not handling the `jitsi_call_started` event, so we'll need to handle this as well here

src/components/Messages/MessageList/livechatCall.js
```diff
@@ -36,10 +36,17 @@ export const CallNotification = (props) => {
 	const [isframe, setIframe] = useState(false);
 	const [show, setShow] = useState(true);
 
-	const acceptClick = () => {
+	const acceptClick = async () => {
 		setShow(!{ show });
 		if (props.rid.t === 'jitsi_call_started') {
-			window.open(`https://meet.jit.si/xlkhe6qe6dyrlttkx${ props.rid.rid }`);
+			const result = await Livechat.videoCall(props.rid);
+			/* result:videoCall:
```
murtaza98, line 43: There's no need to add this comment here.

src/components/Messages/MessageList/livechatCall.js
```diff
@@ -36,10 +36,17 @@ export const CallNotification = (props) => {
 	const [isframe, setIframe] = useState(false);
 	const [show, setShow] = useState(true);
 
-	const acceptClick = () => {
+	const acceptClick = async () => {
 		setShow(!{ show });
 		if (props.rid.t === 'jitsi_call_started') {
-			window.open(`https://meet.jit.si/xlkhe6qe6dyrlttkx${ props.rid.rid }`);
+			const result = await Livechat.videoCall(props.rid);
```
murtaza98, line 42: We might need to do some error handling here, incase the `Livechat.videoCall` fails

src/components/Messages/MessageList/livechatCall.js
```diff
@@ -0,0 +1,73 @@
...
+import PhoneDecline from '../../../icons/phoneOff.svg';
+import { Avatar } from '../../Avatar';
+import { Button } from '../../Button';
+import { Screen } from '../../Screen';
+import { createClassName, getAvatarUrl } from '../../helpers';
+import { parseDate } from '../MessageTime/index';
+import styles from './styles.scss';
+
+
+const DisplayCallIframe = (rid) => (
```
murtaza98, line 16: Do we still need this component? I believe this isn't required now since we're opening jitsi on a new tab right?

src/components/Messages/MessageList/livechatCall.js
```diff
@@ -0,0 +1,73 @@
...
+	<Screen.Content nopadding>
+		<div className={createClassName(styles, 'iframe')}>
+			<iframe style='height:100%' src={`${ Livechat.client.host }/meet/${ rid.rid }`} />
+		</div>
+	</Screen.Content>
+);
+
+
+export const ShowCallTime = (props) => (
```
murtaza98, line 25: I believe this component will no longer be required once [this](https://github.com/RocketChat/Rocket.Chat.Livechat/pull/617/files#r669399143) comment is resolved

src/components/Messages/MessageList/livechatCall.js
Deepak-learner: For our webrtc meet app, we need to display it in an iframe, so we will be needing this component.

Deepak-learner:
PR at rc core => Livechat-jitsi-call => https://github.com/RocketChat/Rocket.Chat/pull/22678

src/components/Messages/Message/index.js
```diff
@@ -89,6 +91,8 @@ const getSystemMessageText = ({ t, conversationFinishedMessage, transferData })
 	|| (t === MESSAGE_TYPE_WELCOME && I18n.t('Welcome'))
 	|| (t === MESSAGE_TYPE_LIVECHAT_CLOSED && (conversationFinishedMessage || I18n.t('Conversation finished')))
 	|| (t === MESSAGE_TYPE_LIVECHAT_STARTED && I18n.t('Chat started'))
+	|| (t === MESSAGE_WEBRTC_CALL && callTimeMessage(callStatus))
```
murtaza98, line 94:
```suggestion
	|| (t === MESSAGE_WEBRTC_CALL && normalizeCallTimeMessage(callStatus))
```

src/components/Messages/Message/index.js
```diff
@@ -89,6 +91,8 @@ const getSystemMessageText = ({ t, conversationFinishedMessage, transferData })
...
+	|| (t === MESSAGE_JITSI_CALL && callTimeMessage(callStatus))
```
murtaza98, line 95: Same as above

src/components/Messages/MessageList/livechatCall.js
murtaza98: Okay makes sense. Lets keep it

src/components/Messages/MessageList/livechatCall.js
```diff
@@ -0,0 +1,61 @@
...
+
+	const acceptClick = async () => {
+		setShow(!{ show });
+		if (props.rid.t === 'jitsi_call_started') {
+			try {
+				const result = await Livechat.videoCall(props.rid);
+				window.open(`https://${ result.videoCall.domain }/${ result.videoCall.room }`);
+				return;
+			} catch (error) {
+				console.error(error);
```
murtaza98, line 36: There's some room for improvement here. Currently we're only logging the error in console which is not very user friendly UX. A good UX would be to show the error message like [this](https://user-images.githubusercontent.com/34130764/125165846-587fcc80-e1b6-11eb-88a4-9736c154ba77.png)

src/routes/Chat/component.js
murtaza98: @Deepak-learner I've resolved this, Please take a look at my solutions and try to understand the changes I've made

murtaza98:
This PR depends on the following PR from core: https://github.com/RocketChat/Rocket.Chat/pull/22690

src/components/helpers.js
```diff
@@ -126,6 +129,20 @@ export const sortArrayByColumn = (array, column, inverted) => array.sort((a, b)
...
 });
 
+export const normalizeCallTimeMessage = (callStatus) => {
+	const timestamp = new Date().toISOString();
+	const time = format(parseISO(timestamp), isToday(parseISO(timestamp)) ? 'HH:mm' : 'dddd HH:mm');
+	if (!callStatus) {
+		return;
+	}
+	if (callStatus === 'accept') {
+		return I18n.t('Call Started at %{time}', { time });
```
murtaza98, line 139: Translations missing in the translation file

src/components/helpers.js
```diff
@@ -126,6 +129,20 @@ export const sortArrayByColumn = (array, column, inverted) => array.sort((a, b)
...
+	}
+	if (callStatus === 'endCall') {
+		return I18n.t('Call Ended at %{time}', { time });
```
murtaza98, line 142: Translations missing in the translation file

Deepak-learner:
@gabriellsh can you please review this pr. Thanks.

murtaza98 (approved):

PR #618, opened 2021-07-04
add joinCallBtn, endCall message, callRing functionalities.

Deepak-learner:
For Jitsi call : If a visitor clicks on the accept button, the visitor will be redirected to a new tab in jitsi call with the agent. As soon as the visitor clicks on the accept button, call start time and Join call button will appear in chat. By mistake if the visitor closes the call window, the visitor has an option to join the call again with the join call button in chat.

For Webrtc call : If visitors click on the accept button, the video call developed with webrtc will begin in iframe. And also call start time and join call button will be there in chat. Along with video calls in the iframe, visitors also have the option to chat with agents.  In the demo implementation for webrtc, right now nothing is there in iframe, but as soon as webrtc part is done at rc core, we can have a video call in iframe.

For both calls, call notification timeout is set to 20 seconds i.e if visitor does not click on either accept or decline button, after 20 seconds call notification will disappear.

Demo jitsi call implementation :
https://drive.google.com/file/d/1QDiyWt9wU7SMcV50opjON_wfDNAG5q6m/view?usp=sharing

Demo webrtc implementation:
https://drive.google.com/file/d/1M_XI-aTuPrUe6AtUul_tkzDurxeSGPGw/view?usp=sharing


src/components/Messages/MessageList/livechatCall.js
```diff
@@ -0,0 +1,83 @@
...
+// join call button function.
+export const ShowJoinCallButton = (props) => (
+	<div className={createClassName(styles, 'joinCallBox')}>
+		<div className={createClassName(styles, 'joinMessage')} >
+			<div className={createClassName(styles, 'videoIcon')} >
+				<VideoIcon width={20} height={20} />
+			</div>
+			{I18n.t('Join my room to start the video call')}
+		</div>
+	    <a className={createClassName(styles, 'joinLink')} rel='noreferrer' target='_blank' href={`https://meet.jit.si/RocketChatXLKhe6QE6dyRLtTkX${ props.roomId.rid }`} > <VideoIcon width={20} height={20} /> {I18n.t('Join Call')} </a>
```
murtaza98, line 45: Why is the following value hard-coded here - `https://meet.jit.si/RocketChatXLKhe6QE6dyRLtTkX` ? The livechat should get the jitsi server info from the core.

src/components/Messages/MessageList/styles.scss
```diff
@@ -33,3 +33,133 @@
...
+	border: 1px solid #e4e7ea;
+}
+
+.joinMessage {
+	display: flex;
+	flex-direction: row;
+
+	padding: 15px;
+
+	font-family: Inter;
```
murtaza98, line 126: Please remove this font-family from here as well to keep the font consistent

src/components/Messages/MessageList/livechatCall.js
```diff
@@ -0,0 +1,83 @@
...
+		<div>
+			{ show ? (<Screen.Content nopadding>
+				<div className={createClassName(styles, 'call')}>
+					<div className={createClassName(styles, 'avatar')}>
+						<Avatar
+							src={getAvatarUrl(props.rid.u.username)}
+						/></div>
+					Incoming video Call
+					<div className={createClassName(styles, 'btn')}>
+						<Button onClick={declineClick} className={createClassName(styles, 'btn1')}> <PhoneDecline width={20} height={20} /> Decline</Button>
```
murtaza98, line 78: Can we use the buttons provided by UI Kit [here](https://rocketchat.github.io/Rocket.Chat.Fuselage/fuselage/master/?path=/story/buttons-button--primary) I believe that's the same components used in the designs [here](https://www.figma.com/file/lsK9hH4EaSWi56DJmcAZk4/Livechat-Video-call-(GSOC)?node-id=2966%3A50)

src/components/Messages/MessageList/livechatCall.js
yokymia: Hello

yokymia: Thank you

src/components/Calls/CallNotification.js
```diff
@@ -0,0 +1,78 @@
...
+									<PhoneAccept width={20} height={20} /><span style='margin-left:5px'> {I18n.t('Accept')} </span>
+								</Button>
+							</div>
+						</div>
+					</div>
+				)
+				: null
+			}
+		    {timeout()}
+			{ isIframe ? <CallIframe rid={rid} /> : null}
```
murtaza98, line 76: Not a good practice to add this Iframe Component inside the `CallNotification`component because both of them are not related. This component should be rendered just like the `CallNotification`component is being rendered [here](https://github.com/RocketChat/Rocket.Chat.Livechat/pull/618/files#diff-68fca7346dd568c2624a0a156ad3ba4b5c82359d4dd691c7d6d51cfc68a239beR150) so we need another variable in the `Store` or perhaps we can use [this](https://github.com/RocketChat/Rocket.Chat.Livechat/pull/618/files#diff-68cfdabf9fa2ce99fa19fdbe25785f2622056c22f88b1f846d3ac7ffffea0744R38) one which I'd created

src/components/Calls/CallNotification.js
```diff
@@ -1,30 +1,36 @@
...
 import { Button } from '../Button';
 import { createClassName, getAvatarUrl } from '../helpers';
+import { CallIframe } from './CallIFrame';
 import styles from './styles.scss';
 
 
-export const CallNotification = ({ callProvider, callerUsername, url, dispatch } = { callProvider: undefined, callerUsername: undefined, dispatch: undefined }) => {
+export const CallNotification = ({ callProvider, callerUsername, url, dispatch, rid } = { callProvider: undefined, callerUsername: undefined, dispatch: undefined, rid: undefined }) => {
 	const [show, setShow] = useState(!!callProvider && !!callerUsername && !!dispatch && !!url);
+	const [isIframe, setIframe] = useState(false);
```
murtaza98, line 18: Again, not a good practise to store the state if the Iframe component within `CallNotification` component

src/components/Calls/CallNotification.js
```diff
@@ -1,30 +1,36 @@
...
+	const { token } = Livechat.credentials;
+	const msg = 'JOIN_CALL_MSG';
+
 
 	const acceptClick = async () => {
 		setShow(!{ show });
-
+		await Livechat.sendMessage({ token, rid, msg });
```
murtaza98, line 25: This isn't a good UI/UX. The message which you're sending here is visible to the agent as well but its of no use to them, see here
image

src/routes/Chat/component.js
```diff
@@ -148,6 +149,7 @@ export default class Chat extends Component {
 			>
 				<Screen.Content nopadding>
 					{ incomingCallAlert && !!incomingCallAlert.show && <CallNotification { ...incomingCallAlert } dispatch={dispatch} />}
+					{ incomingCallAlert && incomingCallAlert.callstatus === 'accept' ? <div> <CallIframe { ...incomingCallAlert } /> </div> : null }
```
murtaza98, line 152: Why surround `CallIFrame` with a new div? I don't  any use of it here. Please correct me if I'm missing something

src/components/Calls/CallNotification.js
```diff
@@ -34,31 +38,38 @@ export const CallNotification = ({ callProvider, callerUsername, url, dispatch }
...
+								<Button onClick={acceptClick} className={createClassName(styles, 'call-notification__content-actions-accept')} >
+									<PhoneAccept width={20} height={20} /><span style='margin-left:5px'> {I18n.t('Accept')} </span>
+								</Button>
+							</div>
 						</div>
 					</div>
 				)
 				: null
 			}
+		    {timeout()}
```
murtaza98, line 73: Instead of calling this function directly, why not use some react events like `componentDidMount()` or something else and call `setTimeout` there

src/components/Messages/Message/index.js
```diff
@@ -91,8 +89,6 @@ const getSystemMessageText = ({ t, conversationFinishedMessage, transferData, ca
 	|| (t === MESSAGE_TYPE_WELCOME && I18n.t('Welcome'))
 	|| (t === MESSAGE_TYPE_LIVECHAT_CLOSED && (conversationFinishedMessage || I18n.t('Conversation finished')))
 	|| (t === MESSAGE_TYPE_LIVECHAT_STARTED && I18n.t('Chat started'))
-	|| (t === MESSAGE_WEBRTC_CALL && normalizeCallTimeMessage(callStatus))
```
murtaza98, line 94: Why are you removing this change we made in our earlier commit :man_shrugging:

src/components/Calls/ShowJoinCallButton.js
```diff
@@ -0,0 +1,43 @@
+import { h } from 'preact';
```
murtaza98, line 1: We ususally don't include a verb within a component name. Perhaps we can simply name this component as `JoinCallButton`

src/components/Calls/CallTime.js
```diff
@@ -0,0 +1,30 @@
...
+	let callDuration = '';
+	if (hh > 0) {
+		callDuration += `${ hh } hours ${ mm } minutes ${ ss } seconds.`;
+	} else if (mm > 0) {
+		callDuration += `${ mm } minutes ${ ss } seconds.`;
+	} else {
+		callDuration += `${ ss } seconds.`;
+	}
+	console.log('zer', callDuration);
+	const time = moment(props.endTime).format('h:mm A');
```
murtaza98, line 22: I believe we already have an package for formatting the date - `date-fns/format'`. So I think we should be using that instead of importing a new package

src/components/Calls/CallNotification.js
```diff
@@ -11,50 +11,53 @@ import { createClassName, getAvatarUrl } from '../helpers';
 import styles from './styles.scss';
 
 
-export const CallNotification = ({ callProvider, callerUsername, url, dispatch } = { callProvider: undefined, callerUsername: undefined, dispatch: undefined }) => {
-	const [show, setShow] = useState(!!callProvider && !!callerUsername && !!dispatch && !!url);
+export const CallNotification = ({ callProvider, callerUsername, url, dispatch, time } = { callProvider: undefined, callerUsername: undefined, dispatch: undefined, time: undefined }) => {
+	const [show, setShow] = useState(true);
+	console.log('asd', show);
```
murtaza98, line 16: Please remove unnecessary logs

src/components/Calls/CallNotification.js
```diff
@@ -11,50 +11,53 @@ import { createClassName, getAvatarUrl } from '../helpers';
...
 	};
 
 	const declineClick = async () => {
 		await dispatch({ incomingCallAlert: null });
+		await dispatch({ ongoingCall: { callStatus: 'decline', time: { time } } });
 	};
 
 	return (
-		<div className={createClassName(styles, 'call-notification')}>
+		<div>
```
murtaza98, line 39: I didn't understand this change
image

I think you're introducing a new parent div, but I'm not sure what's the use of it. Could you please explain its usage incase I've missed something?

src/components/Calls/CallTime.js
```diff
@@ -0,0 +1,30 @@
...
+	const ss = parseInt(Math.abs(props.endTime - props.time) / 1000) % 60;
+	let callDuration = '';
+	if (hh > 0) {
+		callDuration += `${ hh } hours ${ mm } minutes ${ ss } seconds.`;
+	} else if (mm > 0) {
+		callDuration += `${ mm } minutes ${ ss } seconds.`;
+	} else {
+		callDuration += `${ ss } seconds.`;
+	}
+	console.log('zer', callDuration);
```
murtaza98, line 21: Please remove unnecessary logs

src/components/Calls/CallTime.js
```diff
@@ -0,0 +1,30 @@
...
+	const time = moment(props.endTime).format('h:mm A');
+	console.log('bdf', time);
```
murtaza98, line 23: Please remove unnecessary logs

src/components/Calls/styles.scss
```diff
@@ -73,6 +68,71 @@
...
+
+	border: 1px solid #e4e7ea;
+
+	&__content {
+		display: flex;
+		flex-direction: row;
+
+		padding: 15px;
+
+		font-size: 15px;
```
murtaza98, line 90: Again, I'm not sure why are keep changing font properties. We need to keep font properties consistent throughout the application

src/components/Messages/MessageList/index.js
```diff
@@ -111,6 +113,21 @@ export class MessageList extends MemoizedComponent {
 			const previousMessage = messages[i - 1];
 			const message = messages[i];
 			const nextMessage = messages[i + 1];
+			const { incomingCallAlert } = store.state;
+			const { ongoingCall } = store.state;
+
+			if (message.msg === 'Join my room to start the video call' && ongoingCall) {
```
murtaza98, line 119: This logic won't work for other translations apart from English. We need to fix it

src/components/Messages/MessageList/index.js
```diff
@@ -111,6 +113,21 @@ export class MessageList extends MemoizedComponent {
...
+				const { url, callProvider, rid } = incomingCallAlert || '';
```
murtaza98, line 120:
```suggestion
				const { url, callProvider, rid } = incomingCallAlert || {};
```

src/lib/room.js
```diff
@@ -33,7 +33,12 @@ export const processCallMessage = async (message) => {
 			show: true,
 			callProvider: message.t,
 			callerUsername: message.u.username,
-			...message.customFields.jitsiCallUrl && { url: message.customFields.jitsiCallUrl },
+			rid: message.rid,
+			time: message.ts,
+		} });
+		await store.setState({ ongoingCall: {
```
murtaza98, line 39: We could use the `store.setState` above to store update this property too.

src/lib/room.js
```diff
@@ -44,16 +49,17 @@ export const processCallMessage = async (message) => {
 
 const processMessage = async (message) => {
 	const { incomingCallAlert } = store.state;
-	if (incomingCallAlert) {
-		// TODO: create a new event to handle the call dismiss event, currently we're just dismissing the call alert if a new message is sent which is not a good solution
+	if (incomingCallAlert && message.endTs) {
+		console.log('hhh', incomingCallAlert);
+		await store.setState({ ongoingCall: { callStatus: 'ended', time: message.ts } });
```
murtaza98, line 54: Same goes here
> We could use the `store.setState` above to store update this property too.

src/components/Calls/CallNotification.js
Deepak-learner: The div with 'call-notification' should be present only when the show value is true, or else when the show value is false then also style with classname 'call-notification' will be applied and that we do not need. So, I replace the div with classname 'call-notification' with new div.

src/components/Calls/CallIFrame.js
```diff
@@ -1,14 +1,18 @@
 import { h } from 'preact';
 
-import { Screen } from '../Screen';
+import { Livechat } from '../../api';
 import { createClassName } from '../helpers';
 import styles from './styles.scss';
 
+const { localStorage } = window;
+const storedState = JSON.parse(localStorage.getItem('store'));
```
murtaza98, line 8: We have a dedicated store class to fetch this info so I think we should be using that
https://github.com/RocketChat/Rocket.Chat.Livechat/blob/develop/src/store/index.js

src/components/Calls/CallIFrame.js
```diff
@@ -1,14 +1,18 @@
...
 
-export const CallIFrame = (url) => (
-	<Screen.Content nopadding>
+
+export const CallIframe = () => {
+	const url = `${ Livechat.client.host }/meet/${ storedState.room._id }?token=${ storedState.token }&id=${ storedState.user._id }`;
```
murtaza98, line 12: JFYI: There will be some changes required here after based on [this](https://github.com/RocketChat/Rocket.Chat/pull/22847#discussion_r683456279) comment. Please coordinate with Dhruv on this one.

src/components/Calls/JoinCallButton.js
```diff
@@ -0,0 +1,46 @@
...
+	const clickJoinCall = async () => {
+		const { alerts } = store.state;
+		switch (props.callProvider) {
+			case constants.jitsiCallStartedMessageType: {
+				window.open(props.url);
+				break;
+			}
+			case constants.webrtcCallStartedMessageType: {
+				// TODO: add webrtc code here
+				window.open(`${ Livechat.client.host }/meet/${ storedState.room._id }?token=${ storedState.token }&id=${ storedState.user._id }`);
```
murtaza98, line 25: Same goes here
https://github.com/RocketChat/Rocket.Chat.Livechat/pull/618/files?short_path=51e4f55#r684332980

src/components/Calls/JoinCallButton.js
```diff
@@ -0,0 +1,46 @@
...
+export const JoinCallButton = (props) => {
+	const clickJoinCall = async () => {
+		const { alerts } = store.state;
+		switch (props.callProvider) {
+			case constants.jitsiCallStartedMessageType: {
+				window.open(props.url);
+				break;
+			}
+			case constants.webrtcCallStartedMessageType: {
+				// TODO: add webrtc code here
```
murtaza98, line 24: Please remove this TODO

murtaza98:
@Deepak-learner There's a yarn.lock file added to this PR which is not required. We need to remove it. Can you please take care of it?

murtaza98:
> @Deepak-learner There's a yarn.lock file added to this PR which is not required. We need to remove it. Can you please take care of it?

@Deepak-learner Is this fixed? because I can still see that this file is present. Please refer to the file difference tab on github

murtaza98:
@Deepak-learner I found this critical issue on this PR. Could you please take a look at it
- Call Alert notification is still visible even when the call was accepted or the call was ended. To reproduce this issue 1. start a call. 2. do not send any messages after call has started. 3. end the call 4. refresh the livechat screen

src/lib/room.js
```diff
@@ -200,8 +204,9 @@ export const loadMessages = async () => {
 		const lastMessage = messages[messages.length - 1];
 		await store.setState({ lastReadMessageId: lastMessage && lastMessage._id });
 
-		// TODO: create a separate event for starting the call and checking if the call is ongoing
-		if (lastMessage.t === constants.webrtcCallStartedMessageType || lastMessage.t === constants.jitsiCallStartedMessageType) {
+		if (lastMessage.endTs && (lastMessage.t === constants.webrtcCallStartedMessageType || lastMessage.t === constants.jitsiCallStartedMessageType)) {
```
murtaza98, line 207:
```suggestion
if (lastMessage.t === constants.webrtcCallStartedMessageType || lastMessage.t === constants.jitsiCallStartedMessageType) {
			if (lastMessage.endTs) {
				await store.setState({ ongoingCall: { callStatus: 'ended', time: lastMessage.ts }, incomingCallAlert: null });
				return;
			}
			await processCallMessage(lastMessage);
		}
```

There seems to be a common condition here, so merging them makes more sense

murtaza98:
@Deepak-learner Overall this PR looks good now. I just suggested a small change. Please let me know once its complete, so I can approve and merge it. Thanks!

murtaza98 (approved):

PR #623, opened 2021-07-18
Update call status

Deepak-learner:


Deepak-learner:
I create function updateCallStatus at sdk to communicate with rc-core. PR => https://github.com/RocketChat/Rocket.Chat.js.SDK/pull/141 

Making post request at rc-core with with accept or decline call status . PR =>  https://github.com/RocketChat/Rocket.Chat/pull/22688

PR #627, opened 2021-08-07
Allow setGuestName and setGuestEmail to update inputs

### LINKED ISSUE #626: setGuestName and setGuestEmail are not updating the inputs
I am noticing that these functions on the latest version doesn't seem to be working.

hanoii:
Fixed #626


hanoii:
I actually found a problem here, while adding this the internal state is always overwritten to the default (could be empty) even while typing something new, thus not triggering validations or anything. I noticed this in particular as I am working in both this and #628

hanoii:
I improved this considerably, I think.

I noticed a bunch of different issues with how different usecases around setting up name and email:
- setting them was not updating inputs, that's the original issue reported at #626 
- With the above sorted, once it was set, adding your own inputs were not triggering validations.

After reading on
- https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops
- https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html

I refactored a bit how the props to state were copied as were as handled. Doing most on the constructor and handling default changed by set* functions on `componentDidUpdate` as the above docs recommended.

belmeopmenieuwesim:
Please merge this. It has been an issue for long now.

Issue #625, opened 2021-08-06
Logout/destroy widget


hanoii:
I think #482  is a valid issue and something that needs to be added. If you are working on a site that has user authentication and you want to be able to login/logout visitors from the chat instance it should provide a way of doing it.

PR #628, opened 2021-08-09
Reset hook to clear internal state and go back to a first loaded state

hanoii:
This attempts to add support for #625.


hanoii:
now that #640 got merged I wonder if this one can also be looked at?

src/store/Store.js
```diff
@@ -14,14 +16,18 @@ export default class Store {
...
+				storedState = JSON.parse(localStorage.getItem(this.localStorageKey));
+			}
 		} catch (e) {
 			storedState = {};
 		} finally {
 			storedState = typeof storedState === 'object' ? storedState : {};
 		}
 
-		this._state = { ...initialState, ...storedState };
+		this._initialState = initialState;
```
murtaza98, line 29: Any reason for this new variable assignment?

murtaza98:
Hi @hanoii Thanks for the contribution :smile: We recently introduced a similar feature that automatically resets the local storage when a chat gets closed https://github.com/RocketChat/Rocket.Chat.Livechat/pull/666. There instead of clearing the local storage (which you're proposing to do on the reset() method), we instead create a new visitor. So I was wondering if we could do something similar here as well... WDYT?

src/store/Store.js
```diff
@@ -47,7 +53,7 @@ export default class Store {
 		});
 
 		window.addEventListener('visibilitychange', () => {
-			!this._state.minimized && !this._state.triggered && parentCall('openWidget');
+			!this._state.minimized && parentCall('openWidget');
```
murtaza98, line 56: Sorry I didn't get this change. Why are we not checking for the `triggered` property now?

renatobecker:
@hanoii have you looked at https://github.com/RocketChat/Rocket.Chat.Livechat/pull/666 ?

PR #629, opened 2021-08-09
[NEW] WebRTC-call in new tab for mobile devices

Deepak-learner:
If an agent calls the visitor and the visitor is accessing the live chat widget from mobile, then after accepting the call, webrtc-call will start in a new tab instead of starting the call in iframe. And also add join call button will be there in  livechat widget, which can be used to join the call, if visitor by mistake close the webrtc-call tab.

### DIFF
--- a/src/components/Calls/CallIFrame.js
+++ b/src/components/Calls/CallIFrame.js
@@ -6,12 +6,19 @@ import { createClassName } from '../helpers';
 import styles from './styles.scss';
 
 
-export const CallIframe = () => {
+export const CallIframe = (props) => {
 	const { token, room } = store.state;
 	const url = `${ Livechat.client.host }/meet/${ room._id }?token=${ token }`;
+	const callInNewTab = async () => {
+		window.open(url);
+		await store.setState({ ongoingCall: { callStatus: 'ongoingCallInNewTab', time: props.time } });
+		await store.setState({ incomingCallAlert: { show: false, callProvider: props.callProvider } });
+	};
 	return (
-		<div className={createClassName(styles, 'call-iframe')}>
-			<iframe className={createClassName(styles, 'call-iframe__content')} allow='camera;microphone' src={url} />
+		<div>
+			{(window.innerWidth <= 800) && (window.innerHeight <= 630)
+				? <div className={createClassName(styles, 'call-iframe')}> <iframe className={createClassName(styles, 'call-iframe__content')} allow='camera;microphone' src={url} />
+				</div> : callInNewTab() }
 		</div>
 	);
 };
--- a/src/components/Calls/CallIFrame.js
+++ b/src/components/Calls/CallIFrame.js
@@ -6,19 +6,13 @@ import { createClassName } from '../helpers';
 import styles from './styles.scss';
 
 
-export const CallIframe = (props) => {
+export const CallIframe = () => {
 	const { token, room } = store.state;
 	const url = `${ Livechat.client.host }/meet/${ room._id }?token=${ token }`;
-	const callInNewTab = async () => {
-		window.open(url);
-		await store.setState({ ongoingCall: { callStatus: 'ongoingCallInNewTab', time: props.time } });
-		await store.setState({ incomingCallAlert: { show: false, callProvider: props.callProvider } });
-	};
+
 	return (
-		<div>
-			{(window.innerWidth <= 800) && (window.innerHeight <= 630)
-				? <div className={createClassName(styles, 'call-iframe')}> <iframe className={createClassName(styles, 'call-iframe__content')} allow='camera;microphone' src={url} />
-				</div> : callInNewTab() }
+		<div className={createClassName(styles, 'call-iframe')}>
+			<iframe className={createClassName(styles, 'call-iframe__content')} allow='camera;microphone' src={url} />
 		</div>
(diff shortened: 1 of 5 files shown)


src/components/Calls/CallIFrame.js
```diff
@@ -6,12 +6,19 @@ import { createClassName } from '../helpers';
...
+	const callInNewTab = async () => {
+		window.open(url);
+		await store.setState({ ongoingCall: { callStatus: 'ongoingCallInNewTab', time: props.time } });
+		await store.setState({ incomingCallAlert: { show: false, callProvider: props.callProvider } });
+	};
 	return (
-		<div className={createClassName(styles, 'call-iframe')}>
-			<iframe className={createClassName(styles, 'call-iframe__content')} allow='camera;microphone' src={url} />
+		<div>
+			{(window.innerWidth <= 800) && (window.innerHeight <= 630)
```
murtaza98, line 19: I don't think its a good idea to implement this logic over here. This component is named `CallIFrame` so it should only be responsible for dealing with the ongoing call in the iframe - it shouldn't be responsible for redirecting the user to another page

src/components/Calls/CallNotification.js
```diff
@@ -15,15 +16,27 @@ import styles from './styles.scss';
...
 		setShow(!{ show });
 		await Livechat.updateCallStatus('inProgress', rid);
+
 		switch (callProvider) {
 			case constants.jitsiCallStartedMessageType: {
 				window.open(url);
 				break;
 			}
 			case constants.webrtcCallStartedMessageType: {
+				if (window.innerWidth <= 800 && window.innerHeight >= 630) {
```
murtaza98, line 36: This seems to be a duplicate code being used in 2 places. So perhaps we can make a helper function in `src/components/helpers.js` named `isMobileDevice()` and call it from both these places

src/lib/room.js
```diff
@@ -201,16 +201,18 @@ export const loadMessages = async () => {
...
 
 	if (messages && messages.length) {
-		const lastMessage = messages[messages.length - 1];
-		await store.setState({ lastReadMessageId: lastMessage && lastMessage._id });
-
-		if (lastMessage.t === constants.webrtcCallStartedMessageType || lastMessage.t === constants.jitsiCallStartedMessageType) {
-			if (lastMessage.endTs) {
-				await store.setState({ ongoingCall: { callStatus: 'ended', time: lastMessage.ts }, incomingCallAlert: null });
-				return;
+		for (let i = 0; i < messages.length; ++i) {
```
murtaza98, line 204: IMO, we should be relying on the status property on room object instead of looping through all the messages - to check if any call is still on-going

Deepak-learner: According to your point, after refresh the call status property should be dependent on room object instead of looping through all the messages which I will be fixing in my new pr( call declined message at livechat widget). @murtaza as this pr is for webrtc call for mobile-device and for mobile device this pr is working good as expected. I think we can move forward.

src/lib/room.js
```diff
@@ -185,11 +185,13 @@ Livechat.onMessage(async (message) => {
 });
 
 export const getGreetingMessages = (messages) => messages && messages.filter((msg) => msg.trigger);
+export const getLatestCallMessage = (messages) => messages && messages.filter((msg) => msg.actionLinks && msg.actionLinks.length === 2);
```
murtaza98, line 188: I didn't understand why we're adding such vague 
```suggestion
export const getLatestCallMessage = (messages) => messages && messages.filter((msg) => (msg.t === 'livechat_webrtc_video_call' || msg.t === 'livechat_video_call') || !msg.endTs).pop();
```
To me this logic makes more sense because we're checking the message type instead of the action links array length. Remember there can be some cases where people would be adding two buttons/actionLinks to their livechat message, and if we keep your logic here - it would treat such messages too as call messages and the logic applied would fail.

src/lib/room.js
```diff
@@ -204,14 +206,20 @@ export const loadMessages = async () => {
...
-				await store.setState({ ongoingCall: { callStatus: 'ended', time: lastMessage.ts }, incomingCallAlert: null });
-				return;
-			}
-			await processCallMessage(lastMessage);
+	const latestCallMessage = getLatestCallMessage(messages);
+	if (callStatus === 'inProgress') {
+		if (!latestCallMessage) {
+			return;
 		}
+		store.setState({ ongoingCall: { callStatus: 'ongoingCallInNewTab', time: latestCallMessage.ts }, incomingCallAlert: { show: false, callProvider: latestCallMessage[0].t } });
```
murtaza98, line 220:
```suggestion
		store.setState({ ongoingCall: { callStatus: 'ongoingCallInNewTab', time: latestCallMessage.ts }, incomingCallAlert: { show: false, callProvider: latestCallMessage.t } });
```

murtaza98: There seems to be some inconsistency here. Does `latestCallMessage` message only contains a message object or an array of messages? In your code, here it seems like it has array of messages and over [here](https://github.com/RocketChat/Rocket.Chat.Livechat/pull/629/files#diff-5c8aa1cf19e9b52055d8dd0c0c24ebff2b3ab3bc7dec5a041ab5ba59d2f13830R222) it seems it is just a message object

PR #631, opened 2021-08-12
[FIX] Iframe overlay

tiagoevanp:
Calculate the size for iframe wrapper element and adjust when triggered messages appear.


MartinSchoeler (changes requested):
The Iframe still occupies more space then it should on bigger screens (1920x1080 27 inch screen)

MartinSchoeler (approved):

PR #633, opened 2021-08-13
[NEW] Handle endCall and expandView control buttons

djcruz93:
https://user-images.githubusercontent.com/51796498/129451569-56b6ac1e-35d9-476b-a9af-e23ffbd63592.mp4

### DIFF
--- a/src/components/Calls/CallIFrame.js
+++ b/src/components/Calls/CallIFrame.js
@@ -9,25 +9,16 @@ import styles from './styles.scss';
 export const CallIframe = () => {
 	const { token, room, incomingCallAlert } = store.state;
 	const url = `${ Livechat.client.host }/meet/${ room._id }?token=${ token }&layout=embedded`;
-	window.handleIframeClose = () => store.setState({ incomingCallAlert: {
-		show: false,
-		callProvider: incomingCallAlert.callProvider,
-		callerUsername: incomingCallAlert.callerUsername,
-		rid: incomingCallAlert.rid,
-		time: incomingCallAlert.time,
-	} });
+	window.handleIframeClose = () => store.setState({ incomingCallAlert: { ...incomingCallAlert, show: false } });
 	window.expandCall = () => {
 		window.open(
 			`${ Livechat.client.host }/meet/${ room._id }?token=${ token }`,
 			room._id,
 		);
-		return store.setState({ incomingCallAlert: {
-			show: false,
-			callProvider: incomingCallAlert.callProvider,
-			callerUsername: incomingCallAlert.callerUsername,
-			rid: incomingCallAlert.rid,
-			time: incomingCallAlert.time,
-		} });
+		return store.setState({
+			incomingCallAlert: { ...incomingCallAlert, show: false },
+			ongoingCall: { ...ongoingCall, callStatus: "ongoingCallInNewTab" },
+		});
 	};
 	return (
 		<div className={createClassName(styles, 'call-iframe')}>
--- a/src/components/Calls/CallIFrame.js
+++ b/src/components/Calls/CallIFrame.js
@@ -9,14 +9,14 @@ import styles from './styles.scss';
 export const CallIframe = () => {
 	const { token, room, incomingCallAlert } = store.state;
 	const url = `${ Livechat.client.host }/meet/${ room._id }?token=${ token }&layout=embedded`;
-	window.handleIframeClose = async () => store.setState({ incomingCallAlert: {
+	window.handleIframeClose = () => store.setState({ incomingCallAlert: {
 		show: false,
 		callProvider: incomingCallAlert.callProvider,
 		callerUsername: incomingCallAlert.callerUsername,
 		rid: incomingCallAlert.rid,
 		time: incomingCallAlert.time,
 	} });
-	window.expandCall = async () => {
+	window.expandCall = () => {
 		window.open(
 			`${ Livechat.client.host }/meet/${ room._id }?token=${ token }`,
 			room._id,
--- a/src/components/Calls/CallIFrame.js
+++ b/src/components/Calls/CallIFrame.js
@@ -7,15 +7,27 @@ import styles from './styles.scss';
 
 
 export const CallIframe = () => {
-	const { token, room } = store.state;
-	const url = `${ Livechat.client.host }/meet/${ room._id }?token=${ token }&layout=embedded`;
-	window.handleIframeClose = () => store.setState({ incomingCallAlert: null });
+	const { token, room, incomingCallAlert } = store.state;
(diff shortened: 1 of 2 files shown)


src/components/Calls/CallIFrame.js
```diff
@@ -7,8 +7,28 @@ import styles from './styles.scss';
 
 
 export const CallIframe = () => {
-	const { token, room } = store.state;
-	const url = `${ Livechat.client.host }/meet/${ room._id }?token=${ token }`;
+	const { token, room, incomingCallAlert } = store.state;
+	const url = `${ Livechat.client.host }meet/${ room._id }?token=${ token }&layout=embedded`;
```
murtaza98, line 11:
```suggestion
	const url = `${ Livechat.client.host }/meet/${ room._id }?token=${ token }&layout=embedded`;
```

src/components/Calls/CallIFrame.js
```diff
@@ -7,8 +7,28 @@ import styles from './styles.scss';
...
+	window.handleIframeClose = async () => store.setState({ incomingCallAlert: {
+		show: false,
+		callProvider: incomingCallAlert.callProvider,
+		callerUsername: incomingCallAlert.callerUsername,
+		rid: incomingCallAlert.rid,
+		time: incomingCallAlert.time,
+	} });
+	window.expandCall = async () => {
+		window.open(
+			`${ Livechat.client.host }meet/${ room._id }?token=${ token }`,
```
murtaza98, line 21:
```suggestion
			`${ Livechat.client.host }/meet/${ room._id }?token=${ token }`,
```

src/components/Calls/CallIFrame.js
djcruz93: @murtaza98 are you sure doing this would resolve? Did you try this? According to my testing experience, I consciously removed the `/` in the URL. It needs to be removed [here](https://github.com/RocketChat/Rocket.Chat.Livechat/blob/c438eb2ee9a2c0d31f82a1bfaa8dd65888378b5d/src/components/Calls/JoinCallButton.js#L23) for it to work.

murtaza98: Yups, This did solve my issue when I was testing directly by installing new widget on core i.e. `{serverUrl}/livechat` route.

Also sidenote: Usually we don't allow any extra slash at the back of any base URL (eg, in `Site_Url` setting on RC) and the livechat was built on this assumption. For more info, see the logic [here](https://github.com/RocketChat/Rocket.Chat.Livechat/blob/4f60b934e367ffb21152c37e2608bbe660e35ea5/src/api.js#L4)

src/components/Calls/CallIFrame.js
```diff
@@ -7,8 +7,28 @@ import styles from './styles.scss';
...
-	const url = `${ Livechat.client.host }/meet/${ room._id }?token=${ token }`;
+	const { token, room, incomingCallAlert } = store.state;
+	const url = `${ Livechat.client.host }/meet/${ room._id }?token=${ token }&layout=embedded`;
+	window.handleIframeClose = async () => store.setState({ incomingCallAlert: {
+		show: false,
+		callProvider: incomingCallAlert.callProvider,
+		callerUsername: incomingCallAlert.callerUsername,
+		rid: incomingCallAlert.rid,
+		time: incomingCallAlert.time,
+	} });
```
murtaza98, line 18: Oopps... I totally forgot about this earlier :see_no_evil: -  We're maintaining another variable called `ongoingCall` here as well, so we perhaps need to update that too, when the call is either closed, or expanded to new window

djcruz93: Update what value in ongoingCall?

murtaza98: @Deepak-learner Can provide much better answer to this, however let me try
1. When call stopped - we reset this property so it would be `undefined`
2. When call is opened in new tab - we add some new status there - `ongoingCallInNewTab` (Checkout Deepaks this PR for more info - https://github.com/RocketChat/Rocket.Chat.Livechat/pull/629/files)

djcruz93: @Deepak-learner could you please confirm both the points? I'll only proceed after your confirmation.

Deepak-learner: For the first point :There is no need to set property ongoingCallStatus undefined here. As even if the call stopped, we are setting callStatus value with ended. 

For the second point: Yes I agree with @murtaza98 . We can update callStatus value with ongoingCallInNewTab.

src/components/Calls/CallIFrame.js
```diff
@@ -7,8 +7,19 @@ import styles from './styles.scss';
...
+	const url = `${ Livechat.client.host }/meet/${ room._id }?token=${ token }&layout=embedded`;
+	window.handleIframeClose = () => store.setState({ incomingCallAlert: { ...incomingCallAlert, show: false } });
+	window.expandCall = () => {
+		window.open(
+			`${ Livechat.client.host }/meet/${ room._id }?token=${ token }`,
+			room._id,
+		);
+		return store.setState({
+			incomingCallAlert: { ...incomingCallAlert, show: false },
+			ongoingCall: { ...ongoingCall, callStatus: "ongoingCallInNewTab" },
```
murtaza98, line 20:
```suggestion
			ongoingCall: { ...ongoingCall, callStatus: 'ongoingCallInNewTab' },
```

src/routes/Chat/component.js
```diff
@@ -150,7 +150,7 @@ export default class Chat extends Component {
 			>
 				<Screen.Content nopadding>
 					{ incomingCallAlert && !!incomingCallAlert.show && <CallNotification { ...incomingCallAlert } dispatch={dispatch} />}
-					{ ongoingCall && ongoingCall.callStatus === 'accept' ? <CallIframe { ...incomingCallAlert } /> : null }
+					{ incomingCallAlert?.show && ongoingCall && ongoingCall.callStatus === 'accept' ? <CallIframe { ...incomingCallAlert } /> : null }
```
murtaza98, line 153: I don't think we'd be needing to check this condition anymore right? .... Since `ongoingCall` property is not updated when call is expanded

djcruz93: But what about when the call stop button is pressed?

murtaza98: Okay that makes sense :+1: We'd still require this condition here

murtaza98 (approved):

murtaza98:
@djcruz93 Some builds are failing here, can you please take a look? Thanks!

djcruz93:
> @djcruz93 Some builds are failing here, can you please take a look? Thanks!

@murtaza98 Please check now.

Issue #144, opened 2019-01-30
update messages


ggazzo:
as far as I know changed messages will be added as new messages (: please fix it

PR #634, opened 2021-08-14
[NEW] Notify agent on call decline

djcruz93:
Earlier on declining the call, the visitor was only updating the status in the database and not notifying the agent due to which the agent on the meet page still waited for the visitor to join after the decline event.

https://user-images.githubusercontent.com/51796498/129457306-f37e7bef-e24b-47dc-afc6-71f6354c0e39.mp4

## Proposed change

https://user-images.githubusercontent.com/51796498/129470189-1a27ff1e-d9ac-4360-92a6-ef5db1fc32eb.mp4

Uses [#144](https://github.com/RocketChat/Rocket.Chat.js.SDK/pull/144)

### DIFF
--- a/src/components/Calls/CallNotification.js
+++ b/src/components/Calls/CallNotification.js
@@ -12,12 +12,12 @@ import { createClassName, getAvatarUrl } from '../helpers';
 import styles from './styles.scss';
 
 
-export const CallNotification = ({ callProvider, callerUsername, url, dispatch, time, rid } = { callProvider: undefined, callerUsername: undefined, dispatch: undefined, time: undefined }) => {
+export const CallNotification = ({ callProvider, callerUsername, url, dispatch, time, rid, callId } = { callProvider: undefined, callerUsername: undefined, dispatch: undefined, time: undefined }) => {
 	const [show, setShow] = useState(true);
 
 	const acceptClick = async () => {
 		setShow(!{ show });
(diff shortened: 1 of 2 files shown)


murtaza98:
I noticed an issue here related to this change. If the call is declined by visitor, we show `Call declined` message to the user, however to the visitor, it still shows `Call Ended at ....` message

djcruz93:
>I noticed an issue here related to this change. If the call is declined by visitor, we show Call declined message to the user, however to the visitor, it still shows Call Ended at .... message

@Deepak-learner Could you please handle this case?

@murtaza98 Since this PR is related to the agent getting notification so that the waiting for visitor screen changes to call declined, I think we can move forward it.

murtaza98:
> > I noticed an issue here related to this change. If the call is declined by visitor, we show Call declined message to the user, however to the visitor, it still shows Call Ended at .... message
> 
> @Deepak-learner Could you please handle this case? You should create a ticket in your dashboard so that you do not lose the track of this bug.
> 
> @murtaza98 Since this PR is related to the agent getting notification so that `waiting for the visitor` screen changes to `call declined`, I think we can move forward.

Alright! Sounds good :+1: From the notification point of view, this PR looks good, so let me know once this PR is rebased with https://github.com/RocketChat/Rocket.Chat.Livechat/pull/633 changes

murtaza98:
@djcruz93 Could you please resolve the conflicts in this PR?

PR #641, opened 2021-08-26
[NEW] add-jitsi-call-support

Deepak-learner:
If agent intiate the jitsi call, visitor will receive call notification with accept and decline option. If agent click on accept button the visitor will be redirected to new tab with jitsi call with agent. And also join call button will be there in chat, which can be use in case visitor by mistake close the jitsi call tab to rejoin the call.


murtaza98:
Closing this PR since all the changes from here have been merged with PR #646

PR #643, opened 2021-09-06
[IMPROVE] Send widget source in room creation

tiagoevanp:


src/lib/room.js
```diff
@@ -195,6 +195,7 @@ export const loadMoreMessages = async () => {
 
 export const defaultRoomParams = () => {
 	const params = {};
+	params.source = { type: 'widget' };
```
ggazzo, line 198:
```suggestion
	const params = {
	 type: 'widget'
	};
```

ggazzo (changes requested):
I dont think we need to inform the source, its quite obvious to me if we create the chat by the api, the internal server code should infer the type, as any other entrypoint should infer their own type, it should not be something open to the client decide with kind of type its

PR #638, opened 2021-08-22
[NEW] Audio and Video calling in Livechat using WebRTC

murtaza98:

PR #646, opened 2021-09-14
[NEW] Audio and Video calling in Livechat using WebRTC

murtaza98:
Derived from PR #638 with some refactoring


src/components/Calls/CallTime.js
```diff
@@ -0,0 +1,31 @@
...
+import isToday from 'date-fns/isToday';
+import { h } from 'preact';
+
+import I18n from '../../i18n';
+import { createClassName } from '../helpers';
+import styles from './styles.scss';
+
+
+export const CallTime = (props) => {
+	const hh = parseInt(Math.abs(props.endTime - props.time) / 36e5);
```
KevLehman, line 12: In which format are we receivig endTime and time? Wouldn't be possible to use a library to do the conversion to "human time" instead of doing it manually?

src/components/Calls/CallNotification.js
```diff
@@ -0,0 +1,82 @@
...
+
+	const callInNewTab = async () => {
+		const { token } = store.state;
+		const url = `${ Livechat.client.host }/meet/${ rid }?token=${ token }`;
+		await dispatch({ ongoingCall: { callStatus: CallStatus.ON_GOING_CALL_IN_NEW_TAB, time: { time } }, incomingCallAlert: { show: false, callProvider } });
+		window.open(url, rid);
+	};
+
+	const acceptClick = async () => {
+		setShow(!{ show });
```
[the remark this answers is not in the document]
murtaza98: The show flag here controls the state of the call notification component. So the logic here is to hide the call notification, once the user clicks on the accept button
image

src/components/Calls/CallTime.js
murtaza98: Both these variables contain an ISO string. Regarding using the library for calculating duration, I think that would be a good idea. So we're now using [this](https://date-fns.org/v2.25.0/docs/formatDistance) method to do this calculation. No additional import was required since we're already using `date-fns` package in this repo

KevLehman (approved):

PR #648, opened 2021-09-22
[FIX] Issues on Custom Livechat messages

murtaza98:
1. Fix **Offline Success Message** not working

- Setting
image

- The above message should appear here on Livechat when the visitor submits the offline form
image

2. Support HTML content on the **Instructions** message setting

- Setting
image

- The above setting would now support rich contents like hyperlinks, new lines, bold, italic, etc using HTML tags. and this content would be visible over here on the offline form
image

### DIFF
--- a/src/routes/LeaveMessage/component.js
+++ b/src/routes/LeaveMessage/component.js
@@ -183,10 +183,10 @@ export default class LeaveMessage extends Component {
 			{...props}
 		>
 			<Screen.Content>
-				<p className={createClassName(styles, 'leave-message__main-message')}>
-					{hasForm ? message || defaultMessage : unavailableMessage || defaultUnavailableMessage}
-				</p>
-
+				<p className={createClassName(styles, 'leave-message__main-message')}
+					// eslint-disable-next-line react/no-danger
+					dangerouslySetInnerHTML={{ __html: hasForm ? message || defaultMessage : unavailableMessage || defaultUnavailableMessage }}
+				/>
 				{hasForm && this.renderForm(this.props, this.state)}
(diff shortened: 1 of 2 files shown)


src/routes/LeaveMessage/component.js
```diff
@@ -183,10 +183,10 @@ export default class LeaveMessage extends Component {
 			{...props}
 		>
 			<Screen.Content>
-				<p className={createClassName(styles, 'leave-message__main-message')}>
-					{hasForm ? message || defaultMessage : unavailableMessage || defaultUnavailableMessage}
-				</p>
-
+				<p className={createClassName(styles, 'leave-message__main-message')}
+					// eslint-disable-next-line react/no-danger
+					dangerouslySetInnerHTML={{ __html: hasForm ? message || defaultMessage : unavailableMessage || defaultUnavailableMessage }}
```
MartinSchoeler, line 188: Since the idea is to enable the admins to use links and rich content on the livechat message, wouldn't it be better to use markdown instead? We do already have a render function here [here](https://github.com/RocketChat/Rocket.Chat.Livechat/blob/develop/src/components/Messages/MessageText/markdown.js#L84-L85)

I think using dangerouslySetInnerHTML should be reserved to extreme cases 🙂

MartinSchoeler (changes requested):

src/routes/LeaveMessage/component.js
murtaza98: Hi Martin, Nice idea :ok_hand: I've made the changes to support only markdown format. Let me know if it looks good now :smile: Thanks!!

PR #651, opened 2021-10-04
Fix 'Hide agent info' not working on chat transfer message

murtaza98:


src/routes/Chat/container.js
```diff
@@ -127,7 +127,7 @@ export class ChatContainer extends Component {
 				Livechat.sendMessage({ msg, token, rid }),
 			]);
 		} catch (error) {
-			const reason = error?.data?.error ?? error.message;
+			const reason = (error && error.data && error.data.error.error) || error.message;
```
KevLehman, line 130: Why was this change needed? 👀

src/entry.js
```diff
@@ -1,9 +1,11 @@
 import { h, render } from 'preact';
 
+import * as AppImport from './index';
+
 let root = document.getElementById('app') || document.body.firstElementChild;
 
 const init = async () => {
-	const { default: App } = await import('./index');
+	const { default: App } = AppImport;
```
KevLehman, line 8: Why was this change needed? 👀

src/routes/Chat/container.js
murtaza98: I was getting an error of `Unexpected token` while running `yarn run I18n` command. It seems it's not liking the typescript usage here :(

src/entry.js
murtaza98: Same here

src/routes/Chat/container.js
murtaza98: Please check the PR description for error logs

src/entry.js
murtaza98: Please check the PR description for error logs

PR #655, opened 2021-10-23
[FIX] Undo changes to CSS

MartinSchoeler:
Redo some changes that should’ve not passed to develop, we shouldn’t modify tags directly, as this can effect other elements


murtaza98 (approved):
Tested the changes and it is working!!

murtaza98 (approved):

PR #663, opened 2021-11-22
Chore: Add PR title check

KevLehman:
Every PR should follow the same naming convention we have on Rocket.Chat's main repo.


cauefcr (approved):
Can confirm they are the same files we use on the main Rocket.Chat repository, so should work on the GitHub workflow, the worst thing i can say is that it's missing a new line at the end of file.

PR #677, opened 2022-01-13
[NEW] Introduce Widget API method to manage Business Units

murtaza98:
Depends on https://github.com/RocketChat/Rocket.Chat/pull/24162

Please note: Since Business Units is an Enterprise feature, hence this new method will only work if you have a valid enterprise license of Rocket.Chat

### DIFF
--- a/src/lib/hooks.js
+++ b/src/lib/hooks.js
@@ -2,7 +2,7 @@ import { Livechat } from '../api';
 import { store } from '../store';
 import CustomFields from './customFields';
 import { setWidgetLanguage } from './locale';
-import { loadConfig } from './main';
+import { loadConfig, clearBusinessUnit, setBusinessUnit } from './main';
 import { parentCall } from './parentCall';
 import { createToken } from './random';
 import Triggers from './triggers';
@@ -78,8 +78,15 @@ const api = {
 		if (existingBusinessUnit === newBusinessUnit) {
 			return;
 		}
-		store.setState({ businessUnit: newBusinessUnit });
-		await loadConfig();
+		await setBusinessUnit(newBusinessUnit);
+	},
+
+	async clearBusinessUnit() {
+		const { businessUnit } = store.state;
+		if (!businessUnit) {
+			return;
+		}
+		await clearBusinessUnit();
 	},
 
 	clearDepartment() {
--- a/src/lib/hooks.js
+++ b/src/lib/hooks.js
@@ -76,7 +76,7 @@ const api = {
 
 		const { businessUnit: existingBusinessUnit } = store.state;
 
-		return existingBusinessUnit !== newBusinessUnit && await updateBusinessUnit(newBusinessUnit);
+		return existingBusinessUnit !== newBusinessUnit && updateBusinessUnit(newBusinessUnit);
 	},
 
 	async clearBusinessUnit() {
--- a/src/lib/hooks.js
+++ b/src/lib/hooks.js
@@ -2,7 +2,7 @@ import { Livechat } from '../api';
 import { store } from '../store';
 import CustomFields from './customFields';
 import { setWidgetLanguage } from './locale';
-import { loadConfig, clearBusinessUnit, setBusinessUnit } from './main';
+import { loadConfig, updateBusinessUnit } from './main';
 import { parentCall } from './parentCall';
 import { createToken } from './random';
 import Triggers from './triggers';
@@ -78,7 +78,7 @@ const api = {
 		if (existingBusinessUnit === newBusinessUnit) {
 			return;
 		}
-		await setBusinessUnit(newBusinessUnit);
+		await updateBusinessUnit(newBusinessUnit);
 	},
 
 	async clearBusinessUnit() {
@@ -86,7 +86,7 @@ const api = {
 		if (!businessUnit) {
 			return;
 		}
-		await clearBusinessUnit();
+		await updateBusinessUnit();
 	},
 
 	clearDepartment() {
--- a/src/lib/hooks.js
+++ b/src/lib/hooks.js
@@ -75,18 +75,13 @@ const api = {
 		}
 
 		const { businessUnit: existingBusinessUnit } = store.state;
-		if (existingBusinessUnit === newBusinessUnit) {
-			return;
-		}
-		await updateBusinessUnit(newBusinessUnit);
+
+		return existingBusinessUnit !== newBusinessUnit && await updateBusinessUnit(newBusinessUnit);
 	},
 
 	async clearBusinessUnit() {
 		const { businessUnit } = store.state;
-		if (!businessUnit) {
-			return;
-		}
-		await updateBusinessUnit();
+		return businessUnit && updateBusinessUnit();
 	},
 
 	clearDepartment() {
--- a/src/lib/hooks.js
+++ b/src/lib/hooks.js
@@ -27,6 +27,15 @@ const updateIframeGuestData = (data) => {
 	createOrUpdateGuest(guestData);
 };
 
+const isBothArraysContentSame = (arr1, arr2) => {
+	if (arr1.length !== arr2.length) {
+		return false;
+	}
+	const intersection = arr1.filter((x) => arr2.includes(x));
+
+	return intersection.length === arr1.length;
+};
+
 const api = {
 	pageVisited(info) {
 		if (info.change === 'url') {
@@ -67,6 +76,19 @@ const api = {
 		const department = (dept && dept._id) || '';
 
 		updateIframeGuestData({ department });
+	},
+
+	async setBusinessUnitIds(value) {
+		if (!value || !value.businessUnits || !Array.isArray(value.businessUnits)) {
+			throw new Error('Error! Invalid business ids');
+		}
+
+		const { config: { businessUnitIds: existingBusinessUnitIds }, config } = store.state;
+		if (existingBusinessUnitIds && isBothArraysContentSame(existingBusinessUnitIds, value.businessUnits)) {
+			return;
+		}
+		store.setState({ config: { ...config, businessUnitIds: value.businessUnits } });
+		await loadConfig();
 	},
 
 	clearDepartment() {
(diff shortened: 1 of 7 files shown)


murtaza98:
Sample usage of this feature:

```
<script>
		(function(w, d, s, u) {
			w.RocketChat = function(c) { w.RocketChat._.push(c) };
			w.RocketChat._ = [];
			w.RocketChat.url = u;
			var h = d.getElementsByTagName(s)[0],
				j = d.createElement(s);
			j.async = true;
			j.src = 'build/rocketchat-livechat.min.js?_=' + Math.random();
			h.parentNode.insertBefore(j, h);
		})(window, document, 'script', 'http://localhost:8080');
		RocketChat(function() {
			this.setBusinessUnitIds({ businessUnits: [ 'MdssCgP9CKxDtZvX3' ]})
		});
	</script>
```

widget-demo.html
```diff
@@ -51,6 +51,15 @@
 			j.src = 'build/rocketchat-livechat.min.js?_=' + Math.random();
 			h.parentNode.insertBefore(j, h);
 		})(window, document, 'script', 'http://localhost:8080');
+		RocketChat(function() {
```
murtaza98, line 54: I know this isn't related to this PR, but still, I've purposefully added to show a demo about how to use the LiveChat widget API's

src/lib/hooks.js
```diff
@@ -69,6 +78,19 @@ const api = {
 		updateIframeGuestData({ department });
 	},
 
+	async setBusinessUnitIds(value) {
```
renatobecker, line 81:
```suggestion
	async setBusinessUnit(value) {
```

Why are you expecting multiple business unit Ids? This is not supposed to happen, we'll support only one single BU (ID or Name) as paramenter.

src/store/index.js
```diff
@@ -13,6 +13,7 @@ export const initialState = () => ({
 		triggers: [],
 		departments: [],
 		resources: {},
+		businessUnitIds: [],
```
renatobecker, line 16: I feel like this is not the right place to store this information. All properties here are supposed to come from the server and this is not gonna happen with `Business Unit`. This new feature is client-side, so my suggestion is to store the `Business Unit` inside the `Iframe` property and read it from there.

BTW: What happens when `clearLocalStorageWhenChatEnded` is called? The `Business Unit` should not be affected.

renatobecker (changes requested):

src/lib/hooks.js
murtaza98: Alright :+1:

src/store/index.js
murtaza98: Yeah, that makes sense. I've moved it to the root level instead now. I couldn't store it on Iframe prop since that is getting reset when `clearLocalStorageWhenChatEnded` is set. Keeping it on a root level now prevents it from getting cleared

src/lib/hooks.js
```diff
@@ -69,6 +69,19 @@ const api = {
...
+
+		const { businessUnit: existingBusinessUnit } = store.state;
+		if (existingBusinessUnit === newBusinessUnit) {
+			return;
+		}
+		store.setState({ businessUnit: newBusinessUnit });
+		await loadConfig();
+	},
+
 	clearDepartment() {
```
renatobecker, line 85: The ability to clear Business Units is missing, a method for that is needed.

renatobecker (changes requested):

src/lib/hooks.js
murtaza98: Good point! added `clearBusinessUnit` method for that

src/lib/hooks.js
```diff
@@ -69,6 +69,19 @@ const api = {
...
+	async setBusinessUnit(newBusinessUnit) {
+		if (!newBusinessUnit || !newBusinessUnit.trim().length) {
+			throw new Error('Error! Invalid business ids');
+		}
+
+		const { businessUnit: existingBusinessUnit } = store.state;
+		if (existingBusinessUnit === newBusinessUnit) {
+			return;
+		}
+		store.setState({ businessUnit: newBusinessUnit });
```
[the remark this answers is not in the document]
murtaza98: Yups that's indeed is a bug. it should be fixed now, please check. Thanks!!

src/lib/main.js
```diff
@@ -7,9 +7,44 @@ import I18n from '../i18n';
 import store from '../store';
 import constants from './constants';
 
+export const clearBusinessUnit = async () => {
+	const {
+		token,
+		config: existingConfig,
+	} = store.state;
+
+	Livechat.credentials.token = token;
```
renatobecker, line 16: Why do we need to reset `Livechat.credentials.token` ?

src/lib/main.js
```diff
@@ -7,9 +7,44 @@ import I18n from '../i18n';
...
+
+	const { departments } = await Livechat.config({
+		token,
+	});
+
+	await store.setState({ config: { ...existingConfig, departments }, businessUnit: undefined });
```
renatobecker, line 22: Why are we keeping `businessUnit` prop as **undefined** rather than removing the prop?

src/lib/main.js
```diff
@@ -7,9 +7,44 @@ import I18n from '../i18n';
...
+	}
+
+	Livechat.credentials.token = token;
+
+	const { departments } = await Livechat.config({
+		token,
+		businessUnit: newBusinessUnit,
+	});
+
+	await store.setState({ config: { ...existingConfig, departments }, businessUnit: newBusinessUnit });
```
renatobecker, line 41: `setBusinessUnit` and `clearBusinessUnit` are almost equal, the only difference is that the first one will pass a parameter. Why don't we create a helper method that expects an optional parameter e call it within both methods? 

This way we can reduce complexity.
```suggestion
	await store.setState({ config: { ...existingConfig, departments }, ...newBusinessUnit && { businessUnit: newBusinessUnit } });
```

renatobecker (changes requested):

src/lib/main.js
murtaza98: Oppps.. sorry. This one was by mistake. It should be removed now

src/lib/main.js
murtaza98: Actually, I went with this since we didn't have any method in the `store` to unset the property. After you pointed this out, I've added a new helper method there for the same task & used it here

src/lib/main.js
murtaza98: Good idea!! Made this change, please check

src/widget.js
```diff
@@ -263,6 +263,14 @@ function setDepartment(department) {
 	callHook('setDepartment', department);
 }
 
+function setBusinessUnit(businessUnits) {
+	callHook('setBusinessUnit', businessUnits);
+}
+
+function clearBusinessUnit() {
```
renatobecker, line 270: I don't know why but the new method is not working in my local env:

Screen Shot 2022-01-19 at 10 26 11 AM

src/widget.js
```diff
@@ -263,6 +263,14 @@ function setDepartment(department) {
 	callHook('setDepartment', department);
 }
 
+function setBusinessUnit(businessUnits) {
```
renatobecker, line 266:
```suggestion
function setBusinessUnit(businessUnit) {
```

renatobecker (changes requested):

src/lib/hooks.js
```diff
@@ -69,6 +69,26 @@ const api = {
...
+	async setBusinessUnit(newBusinessUnit) {
+		if (!newBusinessUnit || !newBusinessUnit.trim().length) {
+			throw new Error('Error! Invalid business ids');
+		}
+
+		const { businessUnit: existingBusinessUnit } = store.state;
+		if (existingBusinessUnit === newBusinessUnit) {
+			return;
+		}
+		await updateBusinessUnit(newBusinessUnit);
```
KevLehman, line 81:
```suggestion

		return existingBusinessUnit !== newBusinessUnit && await updateBusinessUnit(newBusinessUnit);
```

src/lib/hooks.js
```diff
@@ -69,6 +69,26 @@ const api = {
 		updateIframeGuestData({ department });
 	},
 
+	async setBusinessUnit(newBusinessUnit) {
+		if (!newBusinessUnit || !newBusinessUnit.trim().length) {
+			throw new Error('Error! Invalid business ids');
```
KevLehman, line 74: Instead of throwing an error, can we just log the error to the console? Because a bad config here will disrupt how end users see the widget 😬 

(Not sure if that's inteded tho)

src/lib/hooks.js
```diff
@@ -69,6 +69,26 @@ const api = {
...
+		}
+		await updateBusinessUnit(newBusinessUnit);
+	},
+
+	async clearBusinessUnit() {
+		const { businessUnit } = store.state;
+		if (!businessUnit) {
+			return;
+		}
+		await updateBusinessUnit();
```
KevLehman, line 89:
```suggestion
		return businessUnit && updateBusinessUnit();
```

KevLehman (changes requested):

src/lib/hooks.js
murtaza98: I don't think this is required since the end-user wouldn't be affected if the business unit is not correct. It will just fall back to show all the departments instead of applying the BU filter. 
But let me know if you still think we should remove it.

src/widget.js
```diff
@@ -263,6 +263,14 @@ function setDepartment(department) {
...
+	callHook('setBusinessUnit', businessUnits);
```
murtaza98, line 267:
```suggestion
	callHook('setBusinessUnit', businessUnit);
```

src/widget.js
murtaza98: I might have a clue about this one. This happened on my local as well & I realised that the changes to this file - `src/widget.js` - doesn't get picked up by live-reload server.

So I had to build the project again using `yarn build && yarn start`. Check if that helps you here

renatobecker: It worked. Thanks!

renatobecker:
Code LGTM!

KevLehman (approved):

PR #688, opened 2022-02-24
[FIX] Making sure the 'hide agent info' hides the agent info even with department change.

cauefcr:
Title says it all.


src/components/helpers.js
```diff
@@ -155,6 +155,10 @@ export const normalizeTransferHistoryMessage = (transferData, sender) => {
 		},
 		department: () => {
 			const to = nextDepartment && nextDepartment.name;
+			if (!sender.username) {
+				const from = 'the agent';
```
murtaza98, line 159: We'd have to handle translations for this string. I'd suggest adding a new translation string altogether, in-case agent info is hidden

src/components/helpers.js
```diff
@@ -155,6 +155,9 @@ export const normalizeTransferHistoryMessage = (transferData, sender) => {
 		},
 		department: () => {
 			const to = nextDepartment && nextDepartment.name;
+			if (!sender.username) {
+				return I18n.t('the agent transferred the chat to the department %{to}', { to });
```
KevLehman, line 159:
```suggestion
				return I18n.t('The agent transferred the chat to the department %{to}', { to });
```

A bit nitpicky but 🙈

MartinSchoeler: Wont this conflict with the key? `the_agent_transferred_the_chat_to_the_department_t_3cbe495c` seems to not have caps

cauefcr: changes done

MartinSchoeler (approved):

KevLehman (approved):
Code-wise looks good. Pls try before merging 👀

PR #695, opened 2022-03-25
[FIX] Fixes broken triggers.

cauefcr:
Fixed conditions so that triggers work again.


src/lib/triggers.js
```diff
@@ -172,9 +168,22 @@ class Triggers {
 							this.fire(trigger);
 						}, parseInt(condition.value, 10) * 1000);
 						break;
+					case 'chat-opened-by-visitor':
+						const openFunc = () => {
+							const { user } = store.state;
+							if (user) {
+								if (trigger.runOnce) { store.off('change', openFunc); }
+								setTimeout(() => {
+									this.fire(trigger).then(console.log).catch(console.error).finally(console.debug);
```
KevLehman, line 177: 🤔

src/routes/TriggerMessage/component.js
```diff
@@ -20,7 +20,7 @@ export default class TriggerMessage extends Component {
 	componentDidUpdate() {
 		let height = 0;
 
-		this.ref.current.base.children.forEach((el) => {
+		[...this.ref.current.base.children].forEach((el) => {
```
KevLehman, line 23: Do you need the spread here? 👀

src/lib/triggers.js
tiagoevanp: console.log is necesseary?

src/routes/TriggerMessage/component.js
cauefcr: yes, forEach is undefined without it for some reason

src/lib/triggers.js
cauefcr: console.log is forgotten code, setTimeout was unfortunately needed for it to work

cauefcr: forget that, it's useless

src/routes/TriggerMessage/component.js
tiagoevanp: so... children is not an iterable array type indeed...

tiagoevanp: @cauefcr is doing almost the same as Array.from(childrenEls), that's enough to me...

KevLehman: Yep, the thing is that ... and .foreach should behave similar for arraylike elements (like children lists) 👀 that's why i was wondering if it was needed haha

tiagoevanp (approved):

PR #701, opened 2022-03-31
[FIX] Prevent html rendering on messages

cauefcr:
```<marquee>text</marquee>```, for example, was being rendered, this PR fixes this.

<script> tags weren't, but it's better to remove all the tags.


tiagoevanp (changes requested):
Hey @cauefcr look at src/components/helpers.js we have a function called parse to help us with this kind of problems... What do you think of up to low level of the tree folders and use on room lib? Maybe think in a proper name to that kind of helper (avoiding generic helpers file as improvement).

tiagoevanp (approved):

PR #710, opened 2022-04-18
[FIX] Sync import for i18next load

tiagoevanp:


src/components/App/index.js
```diff
@@ -1,11 +1,18 @@
...
-import '../../i18next';
 import { Provider as StoreProvider, Consumer as StoreConsumer } from '../../store';
 import App from './App';
 
-const AppConnector = () => (
-	<div id='app'>
+const loadI18nextLib = async () => {
+	await import('../../i18next');
+
+	return 'ready';
```
KevLehman, line 9: Do we need the "ready" return or we can just return nothing? 🙈

KevLehman (approved):

KevLehman (dismissed):

KevLehman (approved):

PR #712, opened 2022-04-19
[FIX] Broken triggers

cauefcr:
Tested all 3 of the triggers, they send messages to the user based on configured values correctly.

### DIFF
--- a/src/lib/triggers.js
+++ b/src/lib/triggers.js
@@ -89,8 +89,8 @@ class Triggers {
 	}
 
 	async fire(trigger) {
-		const { token, user, firedTriggers = [] } = store.state;
-		if (!this._enabled || !user) {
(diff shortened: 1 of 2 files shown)


src/lib/triggers.js
```diff
@@ -169,7 +165,7 @@ class Triggers {
 						const openFunc = () => {
 							const { user } = store.state;
 							if (user) {
-								if (trigger.runOnce) { store.off('change', openFunc); }
+								store.off('change', openFunc);
```
KevLehman, line 168: Why removing the 'change' handler? 👀 (not familiar with the triggers code, just wanted to understand how this affects)

cauefcr: Well, without this comments were being sent in an infinite loop :eyes:

tiagoevanp (approved):

PR #713, opened 2022-04-20
[FIX] Translations sync with admin config

tiagoevanp:


src/lib/connection.js
```diff
@@ -14,6 +14,10 @@ let initiated = false;
 const { livechatDisconnectedAlertId, livechatConnectedAlertId } = constants;
 const removeListener = (l) => l.stop();
 
+const loadI18nextLib = async () => {
+	await import('../i18next');
+};
+
```
ggazzo, line 20:
```suggestion
```

src/lib/connection.js
```diff
@@ -29,6 +33,7 @@ const Connection = {
 		try {
 			this.clearListeners();
 			await loadConfig();
+			await loadI18nextLib();
```
ggazzo, line 36:
```suggestion
			await import('../i18next');
```

ggazzo (changes requested):

ggazzo (approved):

PR #715, opened 2022-04-25
[FIX] UI/UX issues on Live Chat widget

MartinSchoeler:
- [x] Quick reply buttons don't fit the UI/Message container
- [ ] Hyperlinks open within the iframe instead of a new browser page
- [ ] Rocket.chat logo is not the latest one


MartinSchoeler:
Moved to https://github.com/RocketChat/Rocket.Chat/pull/25407

PR #717, opened 2022-05-03
[IMPROVE] Deal with new backend state on email validation error

cauefcr:


src/lib/hooks.js
```diff
@@ -11,7 +11,7 @@ import Triggers from './triggers';
 const createOrUpdateGuest = async (guest) => {
 	const { token } = guest;
 	token && await store.setState({ token });
-	const user = await Livechat.grantVisitor({ visitor: { ...guest } });
+	const user = await Livechat.grantVisitor({ visitor: { ...guest } }).catch(() => null);
```
murtaza98, line 14: Wouldn't this ignore any error thrown by the API? Feel we should at least log it.

murtaza98: One more suggestion: instead of .catch() pipeline, I'd recommend using try/catch block since it not only improves code readability but also preserves the parent call context so you won't have to use .bind() method to access any variables in parent function

cauefcr: this will make the user become null, it's used after to show the error message.

KevLehman: I'd also prefer to go with the try catch as it's more explicit 👀

src/lib/hooks.js
```diff
@@ -137,7 +137,9 @@ const api = {
 			api.setDepartment(data.department);
 		}
 
-		createOrUpdateGuest(data);
+		createOrUpdateGuest(data).catch((e) => {
```
KevLehman, line 140: When this fails, are we good just loggin the message to the console? 👀 or should we put an indicator on the ui too? (not sure if this happens already)

src/routes/Chat/container.js
```diff
@@ -57,7 +57,7 @@ class ChatContainer extends Component {
 		}
 
 		const visitor = { token, ...guest };
-		const newUser = await Livechat.grantVisitor({ visitor });
+		const newUser = await Livechat.grantVisitor({ visitor }).catch(() => null);
```
KevLehman, line 60: Same for the try/catch here

src/routes/Register/container.js
```diff
@@ -41,7 +41,7 @@ export class RegisterContainer extends Component {
 
 		await dispatch({ loading: true, department });
 		try {
-			const user = await Livechat.grantVisitor({ visitor: { ...fields, token } });
+			const user = await Livechat.grantVisitor({ visitor: { ...fields, token } }).catch(() => null);
```
KevLehman, line 44: you're already on a try/catch here 👀

src/routes/SwitchDepartment/container.js
```diff
@@ -28,7 +28,7 @@ class SwitchDepartmentContainer extends Component {
 		}
 
 		if (!room) {
-			const user = await Livechat.grantVisitor({ visitor: { department, token } });
+			const user = await Livechat.grantVisitor({ visitor: { department, token } }).catch(() => null);
```
KevLehman, line 31: 😬

KevLehman (changes requested):

MartinSchoeler:
Moved to https://github.com/RocketChat/Rocket.Chat/pull/25411/files

